PR #139, opened 2020-10-01
feat(components/select/select.tsx): add a method to set selected valu…

### LINKED ISSUE #132: [Enhancement] - Set Values to fw-select with mulitple option.
**Description of the enhancement:**

Adding an additional method to the `<fw-select> </fw-select>` component. In order to set values for `fw-select` component with `multiple` property, we need a out of box method from FW. Currently, we are manipulating the value of the select element using array methods and re-rendering the `fw-select` component to DOM.

**Describe the solution you'd like**

We would like to have a method to set values to `fw-select` component. Like `getSelectedItem() => Promise<any>` available for extract values.

narendran-kannan:
…es in fw-select component

Adding an additional method to the fw-select component to set values for the select component with
multiple property.

fix #132

## Checklist:

## How Has This Been Tested?
 Added test for the method.

 n/a

### DIFF
--- a/src/components/select/select.tsx
+++ b/src/components/select/select.tsx
@@ -242,6 +242,16 @@ export class Select {
     return this.options.filter(option => option.selected);
   }
 
+  @Method()
+  setSelectedValues(values: string[]) {
+    if (this.multiple) {
+      this.options.forEach(option => {
+        option.selected = values.includes(option.value);
+      });
+      this.renderInput();
+    }
+  }
+
   render() {
     return (
       <Host
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -324,6 +324,7 @@ export namespace Components {
           * Specifies the select field as a mandatory field and displays an asterisk next to the label. If the attribute’s value is undefined, the value is set to false.
          */
         "required": boolean;
+        "setSelectedValues": (values: string[]) => Promise<void>;
         /**
           * Theme based on which the list box is styled.
          */
--- a/src/components/select/readme.md
+++ b/src/components/select/readme.md
@@ -61,6 +61,16 @@ Type: `Promise<any>`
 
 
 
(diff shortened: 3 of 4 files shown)


narendran-kannan:
@asifahmedfw I've few clarifications regarding this PR, whom I should reach out?

asifahmedfw:
@narendran-kannan You can add your doubts here.

narendran-kannan:
Okay sure @asifahmedfw , who will review and I want to know whether I should add mini-test for this method?

asifahmedfw:
@narendran-kannan please add tests [here](https://github.com/freshworks/crayons/blob/next/src/components/select/select.e2e.ts)

src/components/select/select.tsx
```diff
@@ -242,6 +242,16 @@ export class Select {
     return this.options.filter(option => option.selected);
   }
 
+  @Method()
+  setSelectedValues(values: string[]) {
+    if (this.multiple) {
```
shravan-balasubramanian, line 247: Hey @narendran-kannan Can we make this method common for all the select and not just multi select ?

narendran-kannan: for other select methods, it should accept a single value, not an array of values? Right?
@shravan-balasubramanian

shravan-balasubramanian: Yes!. I have a doubt 
Isn't `document.querySelector('fw-select').value = ['a', 'b', 'c']` working ?
If no, can you make that work instead of another method? thoughts?

narendran-kannan: I can't get this? @shravan-balasubramanian

narendran-kannan: Yes, document.querySelector('fw-select').value = ['1', '2', '3'] is working! @shravan-balasubramanian

shravan-balasubramanian (approved):

PR #167, opened 2020-12-21
feat(select.tsx): added `max` for multselect

### LINKED ISSUE #156: [Enhancement] - add maximumSelectionLength on multiselect select box
**Description of the enhancement:**
Currently, We are not able to set a maximum selection length on the multi-select element.

**Describe the solution you'd like**
need an attribute to set maximum selection length to restrict the selection on a multi-select element
by default, it is unlimited if we set that attribute then it should be restricted to the selection

asifahmedfw:
Added a property to disable the dropdown options once a maximum number of options are selected based
on the `max` property.

fix #156

## Checklist:

## How Has This Been Tested?
Storybook.


src/components/select/select.tsx
```diff
@@ -205,7 +209,7 @@ export class Select {
       (<fw-select-option
         value={option.value}
         selected={option.selected}
-        disabled={option.disabled}
+        disabled={option.disabled || (this.value?.length >= this.max)}
```
shravan-balasubramanian, line 212: Should we disable `fw-select` or `fw-select-option` @asifahmedfw ?

asifahmedfw: If you disable the entire component, you can't remove the items

shravan-balasubramanian (approved):

PR #215, opened 2021-10-01
Crayons checkbox

kulkarni-kashyap:
## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] My commits have standard messages as mentioned in [Contributing Guidelines](./../blob/next/CONTRIBUTING.md)

PR #216, opened 2021-10-01
add keyboard nav support and a11y support

kulkarni-kashyap:
## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/radio-group/radio-group.tsx
+++ b/packages/crayons-core/src/components/radio-group/radio-group.tsx
@@ -7,6 +7,7 @@ import {
   Prop,
   Watch,
   h,
+  Listen
 } from '@stencil/core';
 
 import {
@@ -23,16 +24,25 @@ export class RadioGroup {
 
   @Element() host!: HTMLElement;
 
+  private selectedIndex = 0;
+  private getRadiosPromise;
+  
   /**
    * If true, a radio group can be saved without selecting any option. If an option is selected, the selection can be cleared. If the attribute’s value is undefined, the value is set to false.
    */
   @Prop() allowEmpty = false;
-
+  /**
+   * Label for the component, that can be used by screen readers.
+   */
+   @Prop() label = '';
   /**
    * Name of the component, saved as part of form data.
    */
   @Prop() name = '';
-
+  /**
+   * Indicates the direction of the radio buttons alignment, defaults to vertical alignment.
+   */
+  @Prop() orientation: 'row' | 'column' = 'column';
   /**
    * Default option that is selected when the radio group is displayed on the interface. Must be a valid value corresponding to the fw-radio components used in the Radio Group.
    */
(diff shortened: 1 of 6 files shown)


packages/crayons-core/src/components/radio/radio.tsx
```diff
@@ -70,7 +70,10 @@ export class Radio {
           checked: true,
         });
       } else {
-        this.fwDeselect.emit();
+        this.fwDeselect.emit({
+          value: this.value,
+          checked: true,
```
manoj-krishnan, line 75: Should `checked` be `false` here?

packages/crayons-core/src/components/radio/radio.tsx
```diff
@@ -100,25 +103,28 @@ export class Radio {
...
         onClick={() => this.toggle()}
-        role='radio'
-        tabIndex='0'
+        role="radio"
+        tabIndex="-1"
+        aria-label={this.label}
+        aria-describedby="description"
         aria-disabled={this.disabled ? 'true' : 'false'}
-        aria-checked={this.checked}
+        aria-checked={`${this.checked}`}
```
manoj-krishnan, line 111: Is this being done to force-convert `this.checked` into a string?

packages/crayons-core/src/components/radio/radio.tsx
```diff
@@ -100,25 +103,28 @@ export class Radio {
...
-          )}
-        </label>
+        >
+          <input type="radio"
+            ref={el => this.radio = el}>
+          </input>
+          <label>
+          {
+              this.label !== ''
+              ? <span class="label-field">{this.label}</span>
```
manoj-krishnan, line 121: Same comment as in `checkbox`. Possible to render as text node (without the `span`) and style this using the `label` tag? (assuming we're adding class only for styling)

manoj-krishnan (changes requested):
Added an initial set of comments. Please check.

packages/crayons-core/src/components/radio/radio.tsx
kulkarni-kashyap: Yes correct Manoj, `checked` should be false. Have addressed in latest commit

packages/crayons-core/src/components/radio/radio.tsx
kulkarni-kashyap: Correct Manoj, this is done to convert to string. Have changed to `this.checked ? 'true': 'false'` to be more clear

packages/crayons-core/src/components/radio/radio.tsx
kulkarni-kashyap: We are using `label-field` class for styling only. I didn't modify since it was already implemented that way. We can change it to as suggested. Have modified the same in the latest commit

manoj-krishnan (changes requested):
Please consider these comments along with all comments made in #215

packages/crayons-core/src/components/radio-group/radio-group.tsx
```diff
@@ -49,7 +59,65 @@ export class RadioGroup {
...
+    let previousSelected = this.selectedIndex;
+    switch (event.code) {
+      case 'ArrowDown':
+      case 'ArrowRight':
+        event.preventDefault();
+        radios[previousSelected].setAttribute('tabindex', '-1');
+        radios[previousSelected].checked = false;
+
+        // set currently selectedIndex using roving tabindex technique
+        if (this.selectedIndex === radios.length - 1) {
```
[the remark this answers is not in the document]
kulkarni-kashyap: Agreed!, this will avoid two if-else groups. Thank you for suggestion. Have updated in latest commit

packages/crayons-core/src/components/radio-group/radio-group.tsx
```diff
@@ -102,32 +170,41 @@ export class RadioGroup {
...
       this.value = undefined;
     }
   }
 
-  private getRadios() {
-    return Promise.all(
-      Array.from(this.host.querySelectorAll('fw-radio')).map((r) =>
-        r.componentOnReady()
-      )
+  private async getRadios() {
```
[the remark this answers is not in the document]
kulkarni-kashyap: Yes Manoj, the option to move fetching radios to a lifecycle method will make code simple. I have moved the radios fetching code within `connectedCallback` as this will be the first one that gets executed when component is inserted in DOM and corresponding changes where `getRadios` was called is also made

packages/crayons-core/src/components/radio-group/radio-group.tsx
```diff
@@ -102,32 +170,41 @@ export class RadioGroup {
...
+    if(this.getRadiosPromise) return this.getRadiosPromise;
```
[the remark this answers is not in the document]
kulkarni-kashyap: Have removed the method in latest commit

packages/crayons-core/src/components/radio/radio.scss
```diff
@@ -59,12 +59,12 @@ input[type='radio'] {
...
     user-select: none;
     cursor: pointer;
-    margin-bottom: 0;
+    margin-bottom: 4px;
     vertical-align: middle;
     font-size: 14px;
     line-height: 20px;
     color: #12344d;
-    font-weight: 500;
+    font-weight: 600;
```
kulkarni-kashyap, line 67: There was no difference mentioned for font-weights for title and sub-title in DSM , using the same values as mentioned for checkbox

manoj-krishnan (changes requested):

PR #223, opened 2021-10-04
[FP-228] - Dropdown spec implementation

kishore-kumar-E3682:
## Enhancements
- Converted `option` from state to prop, so that an SelectOptions[] can be passed as input.
- Added `searchable` prop to enable search option in single and multiselect.
- Added `search` prop to pass custom search function to the component.
- In case there is no items or options available for the select  a hint/text  "No Items available" is displayed.

## Bug Fixes
- In case of multi-select, when entered search value is not available in the select-options. The entered value is not removed from the select on foucsing out of the element.

## Keyboard Navigation
- The dropdown is now accessible via arrow keys and pressing the enter button will select / deselect the option.

## Checklist:


kishore-kumar-E3682:
Implemented the changes as part of the PR's - https://github.com/freshworks/crayons/pull/316 , https://github.com/freshworks/crayons/pull/275

PR #305, opened 2021-11-24
Feat | fw-icon Optimisation

detiwari003:
## Checklist:


packages/crayons-core/src/components/icon/icon.scss
```diff
@@ -1,21 +1,12 @@
...
-}
-
 .icon {
-  display: block;
-  color: var(--icon-color, $color-elephant-900);
-
-  height: var(--icon-size, 12px);
-  width: var(--icon-size, 12px);
+  display: inline-block;
+  color: var(--icon-color, var(--color-elephant-900));
```
arvindanta, line 7: color-elephant-900 is a scss variable. Please change it to scss variable

arvindanta (changes requested):

packages/crayons-core/package.json
```diff
@@ -98,9 +97,12 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@freshworks/crayons-icon": "file:../crayons-icon",
     "@popperjs/core": "^2.10.2",
     "@stencil/core": "2.9.0",
-    "moment-mini": "^2.24.0"
+    "moment-mini": "^2.24.0",
+    "hast-util-to-html": "^8.0.3",
```
arvindanta, line 104: These imports are not used in this package. Kindly move them to the correct package where they are being used.

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,155 @@
...
+   */
+  @Prop() height = undefined;
   /**
    * Color in which the icon is displayed, specified as a standard CSS color or as a HEX code.
    */
   @Prop() color = '';
+  /**
+   * Enable Intersection Observer. Default is true.
+   */
+  @Prop() lazy = 'true';
```
arvindanta, line 49: Please make this prop as boolean

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,155 @@
...
+    if (lazy === 'false') this.eval_xOb = true;
+    else if (lazy === 'true' || lazy === undefined) {
+      this.eval_xOb = visible;
+    }
+    if (!Build.isBrowser || !name || !this.eval_xOb) {
+      return;
+    }
+
+    if (
+      name.indexOf('svg') > -1 &&
```
arvindanta, line 80: Kindly check if all 3 conditions are required. Can It be accepted via a specific prop instead if poss ?

packages/crayons-core/src/components/icon/readme.md
```diff
@@ -1,13 +1,23 @@
 # Icon (fw-icon)
-fw-icon displays an icon-sized image that imparts meaning to the component it is associated with.
+fw-icon is a renderer of an svg file which displays an icon-sized image that imparts meaning to the component it is associated with.
+We have covered following topics in the implementation of fw-icon as a Icon Library.
 
-## Demo
+1. Enable fw-icon as an optimized renderer for a SVG with inbuilt capabilities like Intersection-Observer and Fetch-API Memoization.
+   Go through the docs to undertand the various props it supports.
+2. React-Support as part of Caryons Release.
```
arvindanta, line 7: Please fix all the typos in this readme

packages/crayons-core/src/components/icon/readme.md
```diff
@@ -30,29 +40,150 @@ function App() {
...
+import ReactDOM from "react-dom";
+import { FwIcon } from "@freshworks/crayons/react";
+import IconLibRegistry from '@freshworks/crayons-icon/icon-utils/icon-registry';
+// Note : Its a sample code and path values can change.
+IconLibRegistry.registerIconLib([
+                                      {lib:'feather', path: 'https://cdn.jsdelivr.net/npm/feather-icons@4.28.0/dist/icons'},
+                                      {lib:'heroicons', path: 'https://cdn.jsdelivr.net/npm/heroicons@0.4.2/outline'},
+                                      {lib:'system', path: 'https://crayons.freshworks.com//custom-icons'},
+                                      // For dev you may point lib 'system' to localhost public custom icon directory
+                                      {lib:'crayons', path: 'https://unpkg.com/@freshworks/crayons@3.0.0-beta.4/dist/crayons/icon-assets/icons'}
```
arvindanta, line 75: Please provide the correct path here

packages/crayons-icon/icon-utils/icon-fetch-memoize.js
```diff
@@ -0,0 +1,93 @@
+import IconLibRegistry from './icon-registry';
+import { parse } from 'svg-parser';
+import { toHtml } from 'hast-util-to-html';
+import { CRAYONS_ICONS_ASSET_PATH } from '@freshworks/crayons-icon';
+
+const iconCache = {};
+const requestCache = {};
+
+export async function fetchIcon(icon, lib) {
+  lib == 'crayons' && !IconLibRegistry.getIconLib(lib)
```
arvindanta, line 10: use === instead of ==

packages/crayons-icon/icon-utils/icon-fetch-memoize.js
```diff
@@ -0,0 +1,93 @@
...
+    ? IconLibRegistry.registerIconLib([
+        { lib: lib, path: CRAYONS_ICONS_ASSET_PATH },
+      ])
+    : false;
+
+  const icon_key = `${lib}-${icon}`;
+
+  if (lib == 'crayons' || lib == 'system') {
```
arvindanta, line 18: use === instead of ==

packages/crayons-icon/icon-utils/icon-registry.js
```diff
@@ -0,0 +1,43 @@
...
+
+  registerIconLib(items) {
+    items.forEach((item) => {
+      this.iconLibRegistry.setItem(item.lib, item.path);
+    });
+  }
+
+  getIconLib(libName) {
+    if (!this.iconLibRegistry.getItem(libName)) {
+      libName.indexOf('-') == 0
```
arvindanta, line 30: use === instead of ==

www/.vuepress/components/ExternalLibIconsShowcase.vue
```diff
@@ -0,0 +1,110 @@
...
+                                      {lib:'feather', path: 'https://cdn.jsdelivr.net/npm/feather-icons@4.28.0/dist/icons'},
+                                      {lib:'heroicons', path: 'https://cdn.jsdelivr.net/npm/heroicons@0.4.2/outline'},
+                                      {lib:'system', path: 'http://127.0.0.1:8081/custom-icons'},
+                                      {lib:'crayons', path: 'https://unpkg.com/@freshworks/crayons@3.0.0-beta.4/dist/crayons/icon-assets/icons'}
+                                    ]);
+    const svgDocEle = '{"svg":{"props":{"viewBox":"0 0 615 84","style":"margin-top:30px;"}},"g":null, "paths":[{"props":null},{"props":{"fill":"black"}}]}';
+    try{
+       IconLibRegistry.mutateSVG('system', 'crayons-logo', svgDocEle);
+    }catch(ex){
+      //console.error(ex);
```
arvindanta, line 59: You can uncomment it or add a proper log message here

www/.vuepress/components/IconTSExportsShowcase.vue
```diff
@@ -0,0 +1,89 @@
...
+      el.value = text;
+      document.body.appendChild(el);
+      el.select();
+      document.execCommand('copy');
+      document.body.removeChild(el);
+      document.getElementById(text).style.display='block';
+      setTimeout(() => document.getElementById(text).style.display='none', 2000);
+    },
+  },
+  mounted() {
```
arvindanta, line 38: remove this unused method

www/.vuepress/components/IconGallery.vue
```diff
@@ -1,35 +1,53 @@
...
     getIconList() {
-      const context = require.context('@icon-assets/icons/', true, /\.svg$/);
-      return context.keys().map((key) => key.substring(2, key.length - 4));
+      const context = require.context('@freshworks/crayons-icon/icons/', true, /\.svg$/);
+      return context.keys().map((key) => {
+        let svgName= key.substring(2, key.length - 4);
+        const svg_name =
+          svgName.indexOf('-') > -1
+            ? svgName.split('-').join('_')
+            : (svgName == 'new' || svgName == 'delete')
```
arvindanta, line 29: use === instead of ==

packages/crayons-core/src/components/icon/readme.md
```diff
@@ -30,29 +40,150 @@ function App() {
...
+
+<code-block title="React:Mutate SVG">
+```jsx
+import React from "react";
+import ReactDOM from "react-dom";
+import { FwIcon } from "@freshworks/crayons/react";
+import IconLibRegistry from '@freshworks/crayons-icon/icon-utils/icon-registry';
+// Note : Its a sample code and path values can change.
+IconLibRegistry.registerIconLib([
+                                  {lib:'system', path: 'https://crayons.freshworks.com/custom-icons'},
```
arvindanta, line 98: please add a generic path .

arvindanta (changes requested):

packages/crayons-icon/icon-utils/icon-registry.js
```diff
@@ -0,0 +1,43 @@
+class IconRegistry {
+  constructor() {
+    if (typeof window !== 'undefined') {
+      this.iconLibRegistry = window.sessionStorage;
```
arvindanta, line 4: Instead of using session storage can we look at using global functions/controllers ?, Using a window object can be a problem when mfe apps use crayons. There could be conflicts in the name of the lib being used and they can get overridden by mistake.

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,155 @@
...
+  @State() private visible = false;
+
+  private intersectionObserver: IntersectionObserver;
 
   @State() svgHTML = '';
 
-  @Watch('name')
-  private setSVGState(iconName: string) {
-    this.svgHTML = icons[iconName];
+  @State() ico_height = undefined;
```
arvindanta, line 63: By default if you don't set any value they are undefined. You don't have set this explicitly.

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,155 @@
...
+import { fetchIcon } from '@freshworks/crayons-icon/icon-utils/icon-fetch-memoize';
 
 @Component({
   tag: 'fw-icon',
   styleUrl: 'icon.scss',
   assetsDirs: ['icon-assets'],
   shadow: true,
 })
+/**
+ * Class of the icon. Suppoprts Intersection Observer, Icon Cache, JS Imports for Tree shaking inline SVGs and SVGs from external CDNs or custom path within project.
```
arvindanta, line 19: Please check the typo

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: closed

packages/crayons-core/src/components/icon/readme.md
detiwari003: closed

packages/crayons-core/src/components/icon/readme.md
detiwari003: closed

packages/crayons-icon/icon-utils/icon-registry.js
detiwari003: closed

www/.vuepress/components/ExternalLibIconsShowcase.vue
detiwari003: closed

www/.vuepress/components/IconTSExportsShowcase.vue
detiwari003: closed

www/.vuepress/components/IconGallery.vue
detiwari003: closed

packages/crayons-core/src/components/icon/readme.md
detiwari003: closed

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,155 @@
...
+          'width': `${this.ico_width}px`,
+        }}
+        innerHTML={this.svgHTML}
+      ></div>
+    );
+  }
 
-    if (this.size !== undefined) {
-      style['--icon-size'] = `${this.size}px`;
+  private waitUntilVisible(callback: () => void): void {
```
parsuram-vijaysankar, line 120: Can we move the intersection observer browser check into a util?

packages/crayons-core/src/components/icon/icon.tsx
parsuram-vijaysankar: I would feel comfortable if this is prop is set to false by default for a period of time until we are sure if the lazy render is without bugs.

Will the lazy render impact accessiblity?

packages/crayons-core/src/components/icon/icon.tsx
parsuram-vijaysankar: Yes, I think keeping a separate prop of setting direct SVG for make sense rather than doubling up the name tag!

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,155 @@
...
+    if (
+      name.indexOf('svg') > -1 &&
+      name.indexOf('viewBox') > -1 &&
+      name.indexOf('fill') > -1
+    )
+      this.svgHTML = this.name;
+    else {
+      this.svgHTML = await fetchIcon(name, library);
+    }
+    this.ico_height = this.height !== undefined ? this.height : this.size;
```
parsuram-vijaysankar, line 84: This will be a problem if we have the --icon-size CSS variable

parsuram-vijaysankar (changes requested):
I was able to do only a first pass review as the PR size If pretty big.

Please go over the comment and address while, we go over other parts of the PR.

packages/crayons-icon/icon-utils/icon-registry.js
detiwari003: As discussed , this is okay implementation from optimisation perspective. We can plan other approaches later as its not very important. 
Singletons/Global Functions exist at the process level. This means that if you have 20 users accessing your site at any given moment, they're using the same singleton object.Sessions exist at the user level. That means you can store data per user, not per process. Developers would register 1/2 lib at max and we are only storing path info. So keeping [key:value] pair in window object gives us a good way to make it accesible to all users and across pages. Also if you refresh the page your globals are maintained. In case of singletons it gets destroyed and recreated. Also doesn't works always. We also have to write a global function to handle such registrations and make it available across components which is maintenance overhead. window.sessionStorage is safe and gives us good apis to accomplish our task. We can proper namespace our libs to avoid collision in decimal cases.
https://stackoverflow.com/questions/2821941/session-vs-singleton-pattern.

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: closed

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: closed

packages/crayons-core/package.json
detiwari003: closed

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: closed

packages/crayons-core/src/components/icon/icon.scss
detiwari003: closed

packages/crayons-icon/icon-utils/icon-registry.js
arvindanta: I don't think if 20 users are accessing the site, they would be sharing the singleton object. This will be per consumer of the crayons. Each consumer of the crayons should get an instance of IconRegistry and it won't be shared b/w the consumers so that this won't be a problem with MFE. But if we go with window and session storage approach, We have to inform the consumers to namespace the keys when they register with the icon lib to avoid values being overridden.

@parsuram-vijaysankar kindly share your views on this.

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,151 @@
...
+   */
+  @Prop() height = undefined;
   /**
    * Color in which the icon is displayed, specified as a standard CSS color or as a HEX code.
    */
   @Prop() color = '';
+  /**
+   * Enable Intersection Observer. Default is true.
+   */
+  @Prop() nolazy = false;
```
parsuram-vijaysankar, line 49: Deepak the spirit of the comment that we gave earlier was to make the mutation-based lazy loading of the icons an opt-in and not to be turned on by default!

It's not to make the boolean value false 🙁

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,151 @@
...
+    else if (!nolazy) {
+      this.eval_xOb = visible;
+    }
+    if (!Build.isBrowser || !name || !this.eval_xOb) {
+      return;
+    }
+
+    if (name.indexOf('</svg>') > -1) this.svgHTML = this.name;
+    else {
+      this.svgHTML = await fetchIcon(name, library);
```
parsuram-vijaysankar, line 82: What would happen if the fetch failed? If it fails do we need to fire any fallback or throw any event?

packages/crayons-core/src/components/icon/readme.md
```diff
@@ -1,18 +1,24 @@
...
+
+1. Enable fw-icon as an optimised renderer for SVG with built-in functions such as Intersection-Observer and Fetch-API Memoization.
+   Go through the docs to understand the various props it supports.
+2. React-Support as part of Caryons Release. 
+
+## Exampes Live
 
 ```html live
-<fw-icon name="add-contact" size="18" color="green"></fw-icon>
+<fw-icon name="Header" size="38" color="blue" nolazy></fw-icon>
```
parsuram-vijaysankar, line 13: This would be a good opportunity to rename the icon `Header` -> `header`

packages/crayons-icon/gulpfile.js
```diff
@@ -33,6 +33,10 @@ const generateIconsExportData = async () => {
   };
   try {
     let indexData = '/* eslint-disable */';
+    indexData =
+      indexData +
+      '\n' +
+      "export const CRAYONS_ICONS_ASSET_PATH = 'https://unpkg.com/@freshworks/crayons-icon@canary/icons';";
```
parsuram-vijaysankar, line 39: So based on this configuration we will need to update this URL to `crayons-icon@canary` => `crayons-icon@next` when we go to production? 

Is there a better way to do this?

parsuram-vijaysankar (changes requested):

packages/crayons-icon/icon-utils/icon-registry.js
parsuram-vijaysankar: Good that you brought the point of an MFE scenario... 
Yes, if two apps use the same name, by mistake, that will be hard to debug the problem. 

@detiwari003 Let's go with the singleton approach in this case... Let's remember that always configurations need to be specific to MFEs and a shared state would just make it a bad architecture!

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,155 @@
...
   @Prop() name: string;
-
   /**
-   * Size of the icon, specified in number of  pixels.
-   * Default value is 12px defined using the --icon-size css variable.
+   * Size of the icon, specified in number of  pixels. This will be square coordinates of (w X h) = size X size
    */
-  @Prop() size: number;
-
+  @Prop() size = 16;
```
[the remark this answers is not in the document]
detiwari003: closed. --icon-size restored and defaults added. New code checked in.

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,155 @@
...
+          'height': ` ${this.ico_height}px`,
+          'width': `${this.ico_width}px`,
+        }}
+        innerHTML={this.svgHTML}
+      ></div>
+    );
+  }
 
-    if (this.size !== undefined) {
-      style['--icon-size'] = `${this.size}px`;
```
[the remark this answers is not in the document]
detiwari003: fixed. tested . Closing.

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: This is very small code with no chances of failing.It cannot be part of separate utils as its a sort of useScreen hook equivalent of react. I have optimized it further. If we don't need it , just pass <fw-icon name="xyz" nolazy />. 
Tested and closing the coment.

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: Fixed : --icon-size defaults | added props : crayon-icon | corrected typos | nolazy prop added.

Also Intersection Observer is showing great results and I have tested it number of times with various settings. There are examples in live in vuepress docs and results can be noted on browser. Many react icon libraries are providing similar implementation via useScreen hook which is using xOb only as implementation.

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,154 @@
...
-  @Watch('name')
-  private setSVGState(iconName: string) {
-    this.svgHTML = icons[iconName];
+  @State() eval_xOb = false;
+
+  @Watch('name') private async setSVGState(): Promise<void> {
+    const { name, visible, library, nolazy, crayonIcon } = this;
+
+    if (nolazy) this.eval_xOb = true;
+    else if (!nolazy) {
```
arvindanta, line 73: we can just use else . I don't think we need else if here since its just a boolean.

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,154 @@
...
   @Prop() name: string;
-
   /**
-   * Size of the icon, specified in number of  pixels.
-   * Default value is 12px defined using the --icon-size css variable.
+   * Identifier of the icon. The attribute’s value must be a valid JS Import Name of the svg in the named export from @freshworks/crayons-icon.
    */
-  @Prop() size: number;
-
+  @Prop() crayonIcon = '';
```
arvindanta, line 29: can we rename it to `crayonsIcon` instead of `crayonIcon`

packages/crayons-icon/icon-utils/icon-fetch-memoize.js
```diff
@@ -0,0 +1,93 @@
...
+        })
+        .catch(() => {
+          return '';
+        });
+    }
+
+    iconCache[icon_key] = await requestCache[icon_key];
+  }
+  if (typeof window !== 'undefined') {
+    window.IconCache = iconCache;
```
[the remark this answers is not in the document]
detiwari003: Any Global object is by default added to window object or its attached to application state. However for storing non-state data like constants , its advisable to use existing APIs which provide better security and reliability. We should not code such global functions when our need is fulfilled by existing sessionStorage provided by ES 2015 ONWARDS.
I request you to reconsider this approach as its useful way of implementing this use-case.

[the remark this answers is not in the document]
detiwari003: Fixed. Closed

packages/crayons-icon/icon-utils/icon-fetch-memoize.js
```diff
@@ -0,0 +1,93 @@
...
+export async function fetchIcon(icon, lib) {
+  lib === 'crayons' && !IconLibRegistry.getIconLib(lib)
+    ? IconLibRegistry.registerIconLib([
+        { lib: lib, path: CRAYONS_ICONS_ASSET_PATH },
+      ])
+    : false;
+
+  const icon_key = `${lib}-${icon}`;
+
+  if (lib === 'crayons' || lib === 'system') {
```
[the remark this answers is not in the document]
detiwari003: Crayons is our library. System refers to anything that developers want to maintain in their repo.

packages/crayons-icon/icon-utils/icon-registry.js
```diff
@@ -0,0 +1,43 @@
...
+        '{"svg":{"props":{"viewBox":"0 0 617 85","style":"padding-right:10px;margin-top:30px;"}},"g":{"props":{"fill":"black"}}, "paths":[{"props":{"fill":"brown"}},{"props":null},....,{"props":null}]}';
+      console.error(
+        `Not a valid SVG JSON . Match the syntax like this : ${svgDocEle}. This is an example. In case of no props pass null against 'svg','g' or 'paths'.e.g. "g":null `
+      );
+      throw ex;
+    }
+  }
+
+  registerIconLib(items) {
+    items.forEach((item) => {
```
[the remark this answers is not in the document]
detiwari003: This is as per what shoelace is doing. I can stop this behaviour. Its useful in cases where our existing unpck url is slow and we want to point to cdn version of crayons. For system libs its desirable as developers need to point to their custom domain or repo.

packages/crayons-icon/icon-utils/icon-fetch-memoize.js
```diff
@@ -0,0 +1,93 @@
...
+      Object.keys(data).forEach(function (key) {
+        svgDoc.children[0].children[0].properties[key] = data[key];
+      });
+    }
+    if (mutateProps.paths !== null) {
+      mutateProps.paths.forEach((path, index) => {
+        const data = path.props;
+        if (data !== null)
+          Object.keys(data).forEach(function (key) {
+            svgDoc.children[0].children[0].children[index].properties[key] =
```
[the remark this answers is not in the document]
detiwari003: I implemented as per HAST docs. Thats the way they represent xml/svg in abstarct html syntax trees. Thats why we need to access the nodes via children property. Its not unsafe as we are not accessing any DOM but a svg representaion as HAST.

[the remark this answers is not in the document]
detiwari003: I will retest this again to see if i can add more null checks to cover border cases as well.

packages/crayons-icon/icon-utils/icon-fetch-memoize.js
```diff
@@ -0,0 +1,93 @@
+import IconLibRegistry from './icon-registry';
+import { parse } from 'svg-parser';
+import { toHtml } from 'hast-util-to-html';
+import { CRAYONS_ICONS_ASSET_PATH } from '@freshworks/crayons-icon';
+
+const iconCache = {};
+const requestCache = {};
+
+export async function fetchIcon(icon, lib) {
```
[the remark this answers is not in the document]
detiwari003: I already made it async and then I am awaiting its response. This was the only optimisation I did. Rest this function is memoized , so its already optimised. I will check once more.

packages/crayons-icon/icon-utils/icon-fetch-memoize.js
```diff
@@ -0,0 +1,93 @@
...
+    if (mutateProps.svg !== null) {
+      const data = mutateProps.svg.props;
+      Object.keys(data).forEach(function (key) {
+        svgDoc.children[0].properties[key] = data[key];
+      });
+    }
+    if (mutateProps.g !== null) {
+      const data = mutateProps.g.props;
+      Object.keys(data).forEach(function (key) {
+        svgDoc.children[0].children[0].properties[key] = data[key];
```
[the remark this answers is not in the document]
detiwari003: I will retest once more.

packages/crayons-icon/icon-utils/icon-registry.js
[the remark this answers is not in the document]
detiwari003: Please let me know the final take here. should i stop registration of crayons lib from outside. In shoelace we can do the same thing via **setBase** API . I checked the docs.

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: Fixed. Closing. Disabled by default.

packages/crayons-core/src/components/icon/readme.md
detiwari003: Done. Closed.

packages/crayons-icon/gulpfile.js
detiwari003: Need to think on it. Published url is not available untill we publish. However users can rgister crayons url from outside as well. But once we publish , we need a PR to update it.

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: fixed. closed.

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: fixed. closed.

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: Fixed. Now if fetch api fails, developers will now see an exception message logged in console with root cause.

packages/crayons-core/src/components/icon/icon.tsx
arvindanta: In the latest pull, I still the prop name as `crayonIcon` instead of `crayonsIcon`

packages/crayons-core/src/components/icon/icon.tsx
arvindanta: can u please remove else if and just use else block. this is not resolved yet

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,154 @@
...
+      console.error(
+        "Please provide valid props 'name' or 'icon'.Check the usage docs."
+      );
+    }
   }
 
-  componentWillLoad() {
-    this.setSVGState(this.name);
+  connectedCallback(): void {
+    this.waitUntilVisible(() => {
```
parsuram-vijaysankar, line 91: This method shouldn't even run if the lazy flag is enabled!

packages/crayons-core/src/components/icon/icon.tsx
parsuram-vijaysankar: Correct my understanding 
nolazy = false would make the lazy load feature enabled by default right?

parsuram-vijaysankar: While you are at it please name the prop as lazy as it was clear to understand!

packages/crayons-core/src/components/icon/icon.tsx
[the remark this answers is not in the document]
parsuram-vijaysankar: This is not resolved!
What was the test case scenario that you used to verify this use case?

packages/crayons-icon/icon-utils/icon-registry.js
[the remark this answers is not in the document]
parsuram-vijaysankar: ideally, we should be pointing to our CDN site once it's live. Let's check with the team on if we can allow override or not!

parsuram-vijaysankar (changes requested):
Some of the resolved messages need to be addressed!

packages/crayons-core/package.json
arvindanta: This is not closed. Can you please fix it?

packages/crayons-icon/icon-utils/icon-fetch-memoize.js
detiwari003: closed

packages/crayons-icon/icon-utils/icon-fetch-memoize.js
detiwari003: fixed

packages/crayons-icon/icon-utils/icon-registry.js
detiwari003: I tried to follow and implemented this model.
[](https://codesandbox.io/s/elegant-swanson-i0g07).
This doesn't works in Crayons system and renders a blank object. Copied same module import concept to registry for state persistence but doesn't works for stencil web components.This same approach is used in Shoelace and seems to works till you don't refresh the page. 
We have no choice but go with window.sessionStorage as in PR . Singleton module imports to access lib registrations in icon component doesn't work because of following reasons:-
1. We are using these imports in .tsx files which are in turn managed by Stencil Compiler. So all its relative imports don't seem to follow JS Module import system and creates new instance each time. Shoelace is using icon.ts file and not icon.tsx so module imports persist across boundaries when setting data via setters.
2. Also the data if shared via singleton across modules seems to persist for js files but for UI States this resets on page refresh. This problem Angular also has even if they are using DI and Injector service. Developers have used localStorage to persist the data on page refresh. Threads available on stackoverflow.

packages/crayons-icon/icon-utils/icon-fetch-memoize.js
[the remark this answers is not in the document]
detiwari003: Done now.

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: changed to lazy.

packages/crayons-core/src/components/icon/icon.tsx
detiwari003: ok. Let me try this out. As of now , on failure event , I can substitute a default icon from crayons system . What icon , I can keep as a default here.?

packages/crayons-core/src/components/icon/icon.tsx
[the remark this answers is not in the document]
detiwari003: Parsu, I removed any hard coded size defaults and now if no size is given it takes from --icon-size variable. 
I have made changes to code now.

arvindanta:
@detiwari003 Please revisit the samples in the doc. Make sure the code samples in the demo and usage section are the same. Kindly test this in a React Typescript app. Typescript app is showing error when importing from the `crayons-icon`

packages/crayons-icon/icon-utils/icon-registry.js
```diff
@@ -0,0 +1,10 @@
+export const registerIconLibrary = (items) => {
+  var elements = document.querySelectorAll('fw-icon');
```
parsuram-vijaysankar, line 2: Why are we doing a querySelector to infer if the lib's used?
Ideally, this should be done at the component level right?

parsuram-vijaysankar (changes requested):

packages/crayons-core/src/components/icon/icon-fetch-memoize.js
```diff
@@ -0,0 +1,39 @@
...
+
+export async function fetchIcon(icon, lib, lib_path) {
+  const icon_key = `${lib}-${icon}`;
+
+  if (iconCache[icon_key]) {
+    return iconCache[icon_key];
+  }
+
+  lib === 'custom'
+    ? (lib_path = `${window.location.origin}/${lib_path}`)
```
parsuram-vijaysankar, line 14: Curious to know in which circumstance we would need to use a relative URL, I mean real-world apps would mostly be publishing to a CDN. 

Also, the library type called "custom" may be confusing to support the use case of relative urls. What if the users map an absolute URL and name the library custom.

packages/crayons-core/src/components/icon/readme.md
```diff
@@ -23,28 +29,142 @@ import ReactDOM from "react-dom";
...
+</code-block>
+
+<code-block title="React:Register Ext IconLib">
+```jsx
+import React from "react";
+import ReactDOM from "react-dom";
+import { FwIcon } from "@freshworks/crayons/react";
+import { registerIconLibrary } from '@freshworks/crayons-icon/icon-utils/icon-registry';
+// Note : Its a sample code and path values can change.
+registerIconLib([
```
parsuram-vijaysankar, line 92: `registerIconLib` This example is wrong!

parsuram-vijaysankar (changes requested):

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -1,51 +1,163 @@
...
+   * Identifier of the icon. The attribute’s value must be a valid JS Import Name of the svg in the named export from @freshworks/crayons-icon.
+   */
+  @Prop() dataSvg = '';
+  /**
+   * Identifier of the icon. The attribute’s value must be a valid cdn path.
+   */
+  @Prop() src: string;
+  /**
+   * Identifier of the icon. The attribute’s value must be a valid cdn path.
+   */
```
arvindanta, line 36: Can you correct the description for this Prop ?

packages/crayons-core/src/components/icon/readme.md
arvindanta: I still see the typo here for Crayons. Can you please check it ?

packages/crayons-icon/icon-utils/icon-registry.js
```diff
@@ -0,0 +1,13 @@
...
+  items.forEach((item) => {
+    elements.forEach((ele) => {
+      if (item.lib !== 'crayons' || item.lib !== 'system')
+        ele.library === item.lib ? (ele.src = item.path) : false;
```
arvindanta, line 6: do we need a conditional operator here ?. can we not just use a short circuit operator  ?

packages/crayons-icon/icon-utils/icon-registry.js
```diff
@@ -0,0 +1,12 @@
+/*global globalThis*/
+export const registerIconLibrary = (items) => {
+  globalThis.iconlibs = items;
```
parsuram-vijaysankar, line 3: @detiwari003 rather than using the window reference why not make this utility part of the crayons-core? 

just like the `requestCache` reference, you can simply map to an internal reference within crayon's scope right!

arvindanta, line 3: I think `globalThis` points to window object . Would advice not going with window object. 
Screenshot 2021-12-06 at 5 31 52 PM

packages/crayons-core/src/utils/index.ts
```diff
@@ -259,3 +259,26 @@ export const cyclicDecrement = (value: number, maxValue: number): number => {
   value--;
   return value < 0 ? maxValue : value;
 };
+
+// Adding Intersection-Observer : fw-icon
+export const waitUntilVisible = (
```
parsuram-vijaysankar, line 264: Did you remove the browser check?

packages/crayons-core/src/components/icon/icon-fetch-memoize.js
parsuram-vijaysankar: @detiwari003 Can you check this comment?

packages/crayons-core/src/components/icon/icon-fetch-memoize.js
```diff
@@ -0,0 +1,40 @@
...
+    : true;
+  const fetch_asset_path =
+    lib !== 'crayons'
+      ? `${lib_path}/${icon}.svg`
+      : `${CRAYONS_ICONS_ASSET_PATH}/${icon}.svg`;
+
+  if (!requestCache[icon_key]) {
+    console.log('fetch_asset_path', fetch_asset_path);
```
parsuram-vijaysankar, line 22: Better to remove this console.log as it is polluting our tests.

parsuram-vijaysankar (changes requested):
@detiwari003 BTW, you need to follow up this PR up with relevant test cases.

packages/crayons-core/src/components/icon/readme.md
```diff
@@ -1,18 +1,24 @@
...
 
-## Demo
+"fw-icon/FwIcon" is a renderer of a svg file that displays an icon-sized image which imparts meaning to the html component it is associated with.
+Following features are available as part of the implementation of fw-icon.
+
+1. Enable fw-icon as an optimised renderer for SVG with built-in functions such as Intersection-Observer and Fetch-API Memoization.
+   Go through the docs to understand the various props it supports.
+2. React-Support as part of Caryons Release. 
+
+## Exampes Live
```
arvindanta, line 10: Please fix the typo

packages/crayons-icon/icon-utils/icon-registry.js
arvindanta: @detiwari003 Please test this in a simple HTML apps via cdn import as well.

packages/crayons-core/src/components/icon/icon-utils.ts
```diff
@@ -0,0 +1,102 @@
...
+  iconlibListElem.style.display = 'none';
+
+  let children = '';
+  for (const key in items) {
+    children = children + `<li id=${key}>${items[key]}</li>`;
+  }
+
+  iconlibListElem.innerHTML = children;
+  const firstBodyEle = document.body.firstChild;
+  firstBodyEle.parentNode.insertBefore(iconlibListElem, firstBodyEle);
```
arvindanta, line 22: Hi, how is this different from using window?. This can also lead to collision of the keys

packages/crayons-core/src/components/icon/icon.tsx
```diff
@@ -11,41 +16,155 @@ import icons from './icon-assets/icons.json';
...
   /**
-   * Size of the icon, specified in number of  pixels.
-   * Default value is 12px defined using the --icon-size css variable.
+   * Identifier of the icon. The attribute’s value must be a valid JS Import Name of the svg in the named export from @freshworks/crayons-icon.
+   */
+  @Prop() dataSvg = '';
+  /**
+   * Valid for fw-icon-type - registry only. This sets a context setter object which can then be used to render related icon from libraries.
+   */
+  @Prop() dataRegisterLibs: any;
```
parsuram-vijaysankar, line 33: Any reason why this is a public prop?
Do you expect developers to set anything in this prop?

parsuram-vijaysankar (changes requested):

parsuram-vijaysankar (changes requested):
@detiwari003 
`CRAYONS_ICONS_ASSET_PATH` config comes in from crayons-icons.

Let's not do this sort of cross-linking, new developer setup get very tricky in if they need to run additional scripts on different packages before crayons-core can work.

PR #332, opened 2021-12-07
feat(data-table): adding DataTable component

Vignesh-Manogar-E3433:
## DataTable component's initial implementation. 
Features that are part of this implementation are:

- Basic styling and layout for data table
- Selectable rows. 
- Sticky header on vertical scroll
- Accessibility added for this set of features

Steps to run vuepress documentation:

- From root of project run npm run docs:dev
- Open the link that shows up in console after running above command.
- Navigate to DataTable component in sidebar.

NOTES:

- Using Table and related tags for creating the layout of the component.
  -  Table element seems to have better browser support. When trying to replicate table component with div tags, faced a couple of issues. Like CSS grid layout having only partial support(older version of specification) in IE. Some important aria tags that are necessary also don't seem to have good browser support. Trying to replicate what table or css grid does through JS would cost us too much later.  Most libraries have gone ahead with using table tags for the same reasons. 
  - https://caniuse.com/css-grid
  - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Gridcell_role#best_practices
  - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Gridcell_role#accessibility_concerns
- Not introducing Table headers/rows as a separate component. Using props to populate children in this component.
  - Table tags don't seem to support Shadow DOM as per specs.
  - https://dom.spec.whatwg.org/#dom-element-attachshadow
  - Table specs have limited permitted content. Don't seem to appreciate non-semantic elements in between table components. 
  - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table
  - All components in Crayons seem to be using ShadowDOM. So didn't want to create components that break the general rule.
  - Table row component might not be used anywhere else outside a table/grid. Components that looks similar are usually  list elements.
- Unlike row and header choice, planning to create pagination as a separate component in crayons. That will be consumed in data-grid.

## Checklist:
 
## How Has This Been Tested?
Manually verified variations in Storybook and Vuepress docs. Added tests for the component.


arvindanta (approved):
LGTM

PR #335, opened 2021-12-08
feat(pagination): compact pagination

kulkarni-kashyap:
## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/pagination/pagination.tsx
+++ b/packages/crayons-core/src/components/pagination/pagination.tsx
@@ -6,7 +6,6 @@ import {
   Event,
   EventEmitter,
   Method,
-  State,
   Watch,
 } from '@stencil/core';
 @Component({
@@ -15,8 +14,8 @@ import {
   shadow: true,
 })
 export class Pagination {
-  @State() end;
-  @State() start;
+  private end;
+  private start;
 
   /**
    * The current page number.
@@ -73,7 +72,7 @@ export class Pagination {
   }
 
   private getStartRecord() {
-    return (this.page - 1) * this.perPage + 1;
+    return Math.max((this.page - 1) * this.perPage + 1, 1);
   }
 
   private getEndRecord() {
@@ -94,29 +93,16 @@ export class Pagination {
   }
 
   private goToPrevious() {
-    this.start = Math.max(this.start - this.perPage, 1);
-    this.end =
-      this.start - this.end !== this.perPage
-        ? this.start + this.perPage - 1
-        : this.end - this.perPage;
-
     this.page = Math.max(1, this.page - 1);
     this.fwChange.emit({
-      previousPage: this.page === 1 ? -1 : this.page - 1,
-      currentPage: this.page,
-      nextPage: this.page + 1,
+      page: this.page,
     });
   }
 
   private goToNext() {
-    this.start =
-      this.end !== this.total ? this.start + this.perPage : this.start;
-    this.end = Math.min(this.end + this.perPage, this.total);
     this.page = Math.min(this.getLastPage(), this.page + 1);
     this.fwChange.emit({
-      previousPage: this.page - 1,
-      currentPage: this.page,
-      nextPage: this.page === this.getLastPage() ? -1 : this.page + 1,
+      page: this.page,
     });
   }
 
--- a/packages/crayons-core/src/components/pagination/pagination.tsx
+++ b/packages/crayons-core/src/components/pagination/pagination.tsx
@@ -64,11 +64,12 @@ export class Pagination {
           <span class='record'>{this.start}</span> to{' '}
           <span class='record'>{this.end}</span> of {this.totalRecords}
         </div>
-        <fw-button-group>
+        <fw-button-group label='Pagination controls'>
           <fw-button
             disabled={this.start === 1}
             color='secondary'
             size='icon'
+            aria-label='Previous'
             onFwClick={() => this.previous()}
           >
             <fw-icon name='chevron-left'></fw-icon>
@@ -77,6 +78,7 @@ export class Pagination {
             disabled={this.end === this.totalRecords}
             color='secondary'
             size='icon'
+            aria-label='Next'
             onFwClick={() => this.next()}
           >
             <fw-icon name='chevron-right'></fw-icon>
--- a/packages/crayons-core/src/components/pagination/pagination.tsx
+++ b/packages/crayons-core/src/components/pagination/pagination.tsx
@@ -6,6 +6,7 @@ import {
   Event,
   EventEmitter,
   Method,
+  State,
 } from '@stencil/core';
 @Component({
   tag: 'fw-pagination',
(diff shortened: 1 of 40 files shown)


packages/crayons-core/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,106 @@
...
+    this.end = this.start + this.recordsPerPage - 1;
+    this.totalRecords = this.totalRecords || this.end;
+  }
+
+  private goToPrevious() {
+    this.start = Math.max(this.start - this.recordsPerPage, 1);
+    this.end =
+      this.start - this.end !== this.recordsPerPage
+        ? this.start + this.recordsPerPage - 1
+        : this.end - this.recordsPerPage;
```
parsuram-vijaysankar, line 64: Setting end and start eagerly without changing page make lead to errors in computation.
What if the pagination is a promise that needs to talk to the back end?

Should we make the internal state react based on the outer components prop set only?

packages/crayons-core/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,106 @@
...
+  componentWillLoad() {
+    // If both totalRecords and page are passed in but if the start record of current page comes out to be
+    // greater than total number of records then assign start to the last page's first record number.
+    this.start = this.totalRecords
+      ? (this.page - 1) * this.recordsPerPage + 1 > this.totalRecords
+        ? this.totalRecords - this.recordsPerPage + 1
+        : (this.page - 1) * this.recordsPerPage + 1
+      : (this.page - 1) * this.recordsPerPage + 1;
+    this.end = this.start + this.recordsPerPage - 1;
+    this.totalRecords = this.totalRecords || this.end;
```
parsuram-vijaysankar, line 56: Let's not mutate `totalRecords` this should be something that the parent component mandatory sets.

packages/crayons-core/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  @Method()
+  async next() {
+    this.goToNext();
+  }
+
+  componentWillLoad() {
+    // If both totalRecords and page are passed in but if the start record of current page comes out to be
+    // greater than total number of records then assign start to the last page's first record number.
+    this.start = this.totalRecords
```
parsuram-vijaysankar, line 50: rather than do this let's make sure `this.page` is pointing to the last offset if it's pointing to an out of bounds pointer.

packages/crayons-core/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  private goToPrevious() {
+    this.start = Math.max(this.start - this.recordsPerPage, 1);
+    this.end =
+      this.start - this.end !== this.recordsPerPage
+        ? this.start + this.recordsPerPage - 1
+        : this.end - this.recordsPerPage;
+
+    this.fwChange.emit({ startRecord: this.start, endRecord: this.end });
```
parsuram-vijaysankar, line 66: We may need to emit the `currentPage` and `requestedPage`
or `currentPage`, `nextPage`, `previousPage`

rather than sending the start and end record. Mostly we would request the backend API to give us the next or previous pages.

packages/crayons-core/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,106 @@
...
+        <div class='current-record'>
+          <span class='record'>{this.start}</span> to{' '}
+          <span class='record'>{this.end}</span> of {this.totalRecords}
+        </div>
+        <fw-button-group label='Pagination controls'>
+          <fw-button
+            disabled={this.start === 1}
+            color='secondary'
+            size='icon'
+            aria-label='Previous'
```
parsuram-vijaysankar, line 88: We need to move all lang keys to props!

parsuram-vijaysankar (changes requested):

packages/crayons-core/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,106 @@
...
+    this.end = Math.min(this.end + this.recordsPerPage, this.totalRecords);
+
+    this.fwChange.emit({ startRecord: this.start, endRecord: this.end });
+  }
+
+  render() {
+    return (
+      <Host>
+        <div class='current-record'>
+          <span class='record'>{this.start}</span> to{' '}
```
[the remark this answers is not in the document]
arvindanta: @parsuram-vijaysankar , here only the texts `to` and `of` have to be localised.
we can use the same decorator approach here for `to` and `of` text with a default value and a key. 
```
@i18n( defaultValue='of' , key='pagination.of')
@State
ofText = ''
```
and this can be used like `{ this.ofText }`

packages/crayons-core/src/components/pagination/pagination.tsx
arvindanta: we can have a mutable prop with i18n decorator for aria-label here .

packages/crayons-core/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,106 @@
...
+   */
+  @Prop() page = 1;
+  /**
+   * The total number of records.
+   */
+  @Prop({ mutable: true }) totalRecords: number;
+  /**
+   *The number of records to be shown per page. Defaults to 10.
+   */
+  @Prop() recordsPerPage = 10;
```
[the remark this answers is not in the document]
kulkarni-kashyap: Accommodated in latest commit

packages/crayons-core/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,106 @@
...
+  @State() start;
+
+  /**
+   * The current page number.
+   */
+  @Prop() page = 1;
+  /**
+   * The total number of records.
+   */
+  @Prop({ mutable: true }) totalRecords: number;
```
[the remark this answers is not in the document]
kulkarni-kashyap: Accommodated in latest commit

packages/crayons-core/src/components/pagination/pagination.tsx
kulkarni-kashyap: Accommodated in latest commit

packages/crayons-core/src/components/pagination/pagination.tsx
kulkarni-kashyap: Agreed Parsu, have accommodated in latest commit

packages/crayons-core/src/components/pagination/pagination.tsx
kulkarni-kashyap: Have made the necessary changes in latest commit by making `total` as mandatory param

packages/crayons-core/src/components/pagination/pagination.tsx
kulkarni-kashyap: Addressed in the latest commit by introducing a `isLoading` bool, which can be set to either true/false depending on whether the records for current page is loading or not. 

The start and end will be reset upon clicking either next/previous buttons and then these will be disabled if isLoading is set to `true` until it will be set to `false` again. This can be used in scenarios where records fetch is async.

parsuram-vijaysankar (approved):
LGTM!

PR #376, opened 2022-01-18
fix(fw-tag): font size and padding fix

kishore-kumar-E3682:
increased font size to 14px and changed padding/margin

## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/tag/tag.scss
+++ b/packages/crayons-core/src/components/tag/tag.scss
@@ -2,10 +2,10 @@
   display: inline-flex;
   align-items: center;
   width: auto;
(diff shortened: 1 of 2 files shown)


kaushal-freshworks (approved):
fw-avatar {
      padding-right: 8px;
    }
Why this change? Apart from this every thing looks good.

kishore-kumar-E3682:
> fw-avatar { padding-right: 8px; } Why this change? Apart from this every thing looks good.

I have replaced the margin 6px with the padding 4 px to the parent container as per DSM but kept the 8px margin-left as  padding right: 8px to provide some gap between the avatar and the span/text next to it.

PR #377, opened 2022-01-18
fix(fw-checkbox): Removed bottom margin for vertical centre align.

kishore-kumar-E3682:
## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/checkbox/checkbox.scss
+++ b/packages/crayons-core/src/components/checkbox/checkbox.scss
@@ -1,10 +1,14 @@
+/**
+  @prop --checkbox-margin-bottom: Bottom Margin of the checkbox.
+*/
+
 :host {
   display: inline-block;
   position: relative;
 }
 
 .checkbox-container {
-  margin-bottom: 4px;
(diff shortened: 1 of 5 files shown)


packages/crayons-core/src/components/checkbox/checkbox.scss
```diff
@@ -1,10 +1,14 @@
...
+*/
+
 :host {
   display: inline-block;
   position: relative;
 }
 
 .checkbox-container {
-  margin-bottom: 4px;
+  margin-bottom: var(--checkbox-margin-bottom, 4px);
```
kulkarni-kashyap, line 11: Kishore, we actually don't need margin-bottom of 4px here, have removed it in the PR for form render. Since you are passing as 0px anyway, I think we can remove this. Even in DSM there is no specification of margin-bottom for checkbox.

kishore-kumar-E3682: Thanks for the confirmation Shravan, I have removed the margin-bottom here.

kulkarni-kashyap (approved):

PR #395, opened 2022-02-03
feat(fw-input): Added slots and modified styles for fw-input based on DSM

kishore-kumar-E3682:
- changed the disabled background-color to #f7f9fa 
- removed the bottom spacing on the input and increase the font-size to 14px
- reduced the size of the close icon to 8px
- changed the font-weight for the disabled state and placeholder
- exposed css variables to change the normal state font color of label and hint
- modified fw-input as a block element so that consuming component can add their margin or padding
- Added slots to create custom input as shown below

Screenshot 2022-02-04 at 4 52 19 PM

## Checklist:

### DIFF
--- a/packages/crayons-core/src/components.d.ts
+++ b/packages/crayons-core/src/components.d.ts
@@ -404,56 +404,6 @@ export namespace Components {
         /**
           * The date/time to format. If not set, the current date and time will be used.
          */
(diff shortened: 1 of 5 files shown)


sangeet-fw (approved):
LGTM

PR #399, opened 2022-02-04
feat(data-table): add table settings option to data table

Vignesh-Manogar-E3433:
## Description: 
Adding table settings feature to Data table component.

#### Abilities of table settings:
- Hide/Show columns.
- Reorder columns.
- Export and import user modified table configuration.

#### Screen capture explaining features:

https://user-images.githubusercontent.com/61269786/152518068-273ef797-b295-42fa-b501-c86037e3c089.mov

#### Steps to run vuepress documentation:

- From root of project run npm run docs:dev
- Open the link that shows up in console after running above command.
- Navigate to DataTable component in sidebar

## Checklist:
 
## How Has This Been Tested?
- Tested when adding documentation.
- Tested in Chrome, Firefox, Safari browsers.


packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -117,36 +163,53 @@ export class DataTable {
...
+
+  tableContainer: HTMLElement = null;
+
   /**
    * componentWillLoad lifecycle event
    */
   componentWillLoad() {
     this.columnOrdering(this.columns);
+    if (localStorage && this.autoSaveSettings) {
+      const tableId = this.el.id ? `-${this.el.id}` : '';
```
arvindanta, line 188: can we set a unique ID if the `id` is not set by the user ?

packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -117,36 +163,53 @@ export class DataTable {
...
+      const tableSettings = localStorage.getItem(`fw-table${tableId}`);
+      tableSettings && this.setTableSettings(JSON.parse(tableSettings));
```
arvindanta, line 190: Please wrap the `JSON.parse` with `try/catch` block.

packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -390,6 +369,318 @@ export class DataTable {
...
+   * getColumnsState function to call when updating a column state.
+   * Updating state using this function can help update the UI easily.
+   * @param state current state to modify
+   * @param stateKey key that can help identify what object in state to identify
+   * @param keyValuePairs key value pair to modify the object
+   * @returns modifies state object
+   */
+  getColumnsState(state, stateKey, keyValuePairs) {
+    let currentElementIndex;
+    const currentElement = state.filter((stateElement, elementIndex) => {
```
arvindanta, line 383: can we use `findIndex` instead ? and set `currentElement = state[currentElementIndex]`

packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -390,6 +369,318 @@ export class DataTable {
...
+    const otherElements = state.filter(
+      (stateElement) => stateElement.key !== stateKey
+    );
+    const preElements = otherElements.slice(0, currentElementIndex);
+    const postElements = otherElements.slice(
+      currentElementIndex,
+      otherElements.length
+    );
+    if (modifiedCurrentElement.position !== currentElement.position) {
+      preElements.map((preElement) => {
```
arvindanta, line 400: it's better to avoid mutating the `preElement`.

arvindanta, line 400: it's better to avoid mutating the `preElement` and `postElement`.

packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -634,9 +1007,9 @@ export class DataTable {
    * @returns {JSX.Element} table header row
    */
   renderTableHeader() {
-    const selectAllChecked = this.rows
-      .map((row) => row.id)
-      .every((rowId) => this.selected.includes(rowId));
+    const selectAllChecked = this.rows.every((row) =>
```
arvindanta, line 1010: The above operation can be a costly one if the no of rows and selected rows are huge in no. 

can we not just compare the length `this.rows.length === this.selected` instead ?.

packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -390,6 +369,318 @@ export class DataTable {
...
+      this.columnsDragSetting = [...this.columnsDragSetting, columnInfo];
+    });
+    this.columns.forEach((column) => {
+      const columnInfo: any = {};
+      columnInfo.key = column.key;
+      columnInfo.text = column.text;
+      columnInfo.hide = this.orderedColumns.filter(
+        (orderedColumn) => orderedColumn.key === column.key
+      )[0].hide;
+      this.columnsHideSetting = [...this.columnsHideSetting, columnInfo];
```
arvindanta, line 488: instead of updating the state on each iteration of the `forEach`, can we use map and return an array -`[columnInfo]` and then merge it with the state like 
`this.columnsHideSetting = [...this.columnsHideSetting, ...columnsInfo]`

packages/crayons-core/src/components/data-table/data-table.tsx
Vignesh-Manogar-E3433: As per our discussion, added error message for when autosave is present and id is missing for the table.

packages/crayons-core/src/components/data-table/data-table.tsx
Vignesh-Manogar-E3433: Changed as suggested.

packages/crayons-core/src/components/data-table/data-table.tsx
Vignesh-Manogar-E3433: preElement, postElement are slices of original param, hence not mutating the original param. Also, changed map to for loop to avoid any confusion.

packages/crayons-core/src/components/data-table/data-table.tsx
Vignesh-Manogar-E3433: As we discussed, this is necessary to maintain selected state between pagination.

packages/crayons-core/src/components/data-table/data-table.tsx
Vignesh-Manogar-E3433: Changed as suggested.

arvindanta (approved):

PR #443, opened 2022-02-28
fix(data-table): adding cell variants as dependency to table

Vignesh-Manogar-E3433:
Issue: 
Cell variant components appear empty in data-table cell in CO UI staging environment.

Fix: 
This was happening due to stencil not picking up cell variants as dependency for data-table as we are loading them dynamically. In the fix, the components are loaded based on variant name. 

## Checklist:
 
## How Has This Been Tested?
Tested by verifying the variants in vuepress docs in chrome browser.

### DIFF
--- a/packages/crayons-core/src/components/data-table/data-table.tsx
+++ b/packages/crayons-core/src/components/data-table/data-table.tsx
@@ -994,11 +994,14 @@ export class DataTable {
    */
   renderPredefinedVariant(columnVariant: string, cellValue: any) {
     let template: JSX.Element;
-    if (PREDEFINED_VARIANTS_META[columnVariant]) {
-      template = this.renderWebComponent(
-        PREDEFINED_VARIANTS_META[columnVariant].componentName,
-        cellValue
-      );
+    if (columnVariant === 'anchor') {
+      template = <fw-custom-cell-anchor {...cellValue}></fw-custom-cell-anchor>;
+    } else if (columnVariant === 'user') {
+      template = <fw-custom-cell-user {...cellValue}></fw-custom-cell-user>;
+    } else if (columnVariant === 'icon') {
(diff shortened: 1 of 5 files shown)


packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -994,11 +994,14 @@ export class DataTable {
    */
   renderPredefinedVariant(columnVariant: string, cellValue: any) {
     let template: JSX.Element;
-    if (PREDEFINED_VARIANTS_META[columnVariant]) {
-      template = this.renderWebComponent(
-        PREDEFINED_VARIANTS_META[columnVariant].componentName,
-        cellValue
-      );
+    if (columnVariant === 'anchor') {
+      template = <fw-custom-cell-anchor {...cellValue}></fw-custom-cell-anchor>;
```
arvindanta, line 998: I see that `slotText` has been used to pass a slot to the custom variants in the implmentation of `renderWebComponent` . Is this being handled here ?

Can you also pls remove `renderWebComponent` method if this is no longer in use.

arvindanta (changes requested):

packages/crayons-core/src/components/data-table/data-table.tsx
Vignesh-Manogar-E3433: Yes, it is handled in case of custom variants. 

renderWebComponents is a function we might definitely be needing later. I would recommend that we retain this function for now.

arvindanta (approved):

packages/crayons-core/src/components/data-table/data-table.tsx
arvindanta: Thanks Vignesh

PR #544, opened 2022-04-05
Feat/theme demo

kishore-kumar-E3682:
## Description:
Moved the colors sass variables to a generic (from `azure-800` to `color-info-800`) which can be overwritten by the css variable `--fw-color-info-800`. Also exposed the typography and padding / height for button and tabs.

## Checklist:


packages/crayons-core/src/components/button/button.scss
```diff
@@ -1,50 +1,58 @@
 /**
   @prop --fw-button-min-width: minimum width for the button
   @prop --fw-button-label-vertical-padding: vertical padding for the button label
+  @prop --fw-button-height: The height of the normal button.
+  @prop --fw-button-small-height: The height of the small button.
+  @prop --fw-button-border-radius: The height of the small button.
```
arvindanta, line 6: if this is for only small button, can we also add `small` to the css variable ?

packages/crayons-core/src/components/button/button.scss
```diff
@@ -1,50 +1,58 @@
...
+  @prop --fw-button-font-weight: The font weight of the button.
+  @prop --fw-button-font-size: The font size of the normal button.
+  @prop --fw-button-small-font-size: The font size of the small button.
```
arvindanta, line 9: should we also provide customisation for other sizes of button like 'icon' , `icon-small` also?

styles/_utils.scss
```diff
@@ -2,6 +2,7 @@
 @import 'mixins/card';
 @import 'mixins/spacing';
 @import 'mixins/border';
+@import './theme.base';
```
arvindanta, line 5: why we need this import here inside utils ?

arvindanta (changes requested):
Can you please add the global variables in the docs ?. Otherwise consumers will be totally unaware of the variables that needs to be used.

styles/_utils.scss
kishore-kumar-E3682: Moved the them.base code to the theme.scss file.

packages/crayons-core/src/components/button/button.scss
kishore-kumar-E3682: Its a typo, modified it.

packages/crayons-core/src/components/button/button.scss
kishore-kumar-E3682: The icons are passed as slots so we can change the icon size from there.

arvindanta (commented):
Can you please update the docs with all the css variables that can be used at the global level in the docs including font size etc.

styles/_theme.scss
```diff
@@ -1,3 +1,93 @@
...
+
+// Primary colors
+$palette-primary-900: var(--fw-palette-primary-900, $color-elephant-900);
+$palette-primary-800: var(--fw-palette-primary-800, $color-elephant-800);
+$palette-primary-700: var(--fw-palette-primary-700, $color-elephant-700);
+$palette-primary-600: var(--fw-palette-primary-50, $color-elephant-600);
+
+// Secondary colors
+$palette-secondary-700: var(--fw-palette-secondary-700, $color-smoke-700);
+$palette-secondary-400: var(--fw-palette-secondary-300, $color-smoke-400);
```
arvindanta, line 12: Please change it to `--fw-palette-secondary-400`

arvindanta (changes requested):

styles/_theme.scss
kishore-kumar-E3682: Updated the css variable

styles/_theme.scss
```diff
@@ -1,3 +1,93 @@
+// Neutral colors
+$palette-neutral-100: var(--fw-palette-neutral-100, $color-milk);
+
+// Primary colors
+$palette-primary-900: var(--fw-palette-primary-900, $color-elephant-900);
+$palette-primary-800: var(--fw-palette-primary-800, $color-elephant-800);
+$palette-primary-700: var(--fw-palette-primary-700, $color-elephant-700);
+$palette-primary-600: var(--fw-palette-primary-50, $color-elephant-600);
```
arvindanta, line 8: can u change this to `--fw-palette-primary-600` pls ?

styles/_theme.scss
```diff
@@ -1,3 +1,93 @@
...
+$palette-danger-50: var(--fw-palette-danger-50, $color-persimmon-50);
+
+// Info colors
+$palette-info-800: var(--fw-palette-info-800, $color-azure-800);
+$palette-info-100: var(--fw-palette-info-100, $color-azure-100);
+$palette-info-50: var(--fw-palette-info-50, $color-azure-50);
+
+// Elevation shadow
+$elevation-level-1: var(
+  --fw-elevation-level-1,
```
arvindanta, line 46: pls add elevation-level as well in the docs

arvindanta (approved):

arvindanta:
Please do not merge to `next` branch until E2E testing is done with the docs site

PR #570, opened 2022-04-22
fix(tooltip): ability to hover over tooltip content

Vignesh-Manogar-E3433:
Issue:
Not able to hover over tooltip content.

https://user-images.githubusercontent.com/61269786/164599586-f14f487a-eb0d-4d9f-b133-ded4b76cb122.mov

Fix: 
Issue was because we had hide on trigger leave. Now, changing mouseleave event to tooltip host and adding time to hide tooltip content so that people can move to tooltip content before the timeout/hide event.

https://user-images.githubusercontent.com/61269786/164599609-057f9a12-6612-4dfb-be7e-85bff238fc2a.mov

## Checklist:
 
## How Has This Been Tested?
Tested in chrome browser.

### DIFF
--- a/packages/crayons-core/src/components/popover/popover.tsx
+++ b/packages/crayons-core/src/components/popover/popover.tsx
@@ -191,15 +191,21 @@ export class Popover {
       trigger.addEventListener('focus', this.show.bind(this));
       trigger.addEventListener('blur', this.hide.bind(this));
       trigger.addEventListener('mouseenter', this.show.bind(this));
-      trigger.addEventListener('mouseleave', (event: any) => {
-        /**
-         * popovercontent might trigger mouseLeave callback of triggerRef during transition.
-         * This condition avoids flicker that might happen because of the mentioned case.
-         */
-        if (
-          event.relatedTarget === null ||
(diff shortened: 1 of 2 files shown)


kishore-kumar-E3682 (approved):

packages/crayons-core/src/components/popover/popover.tsx
```diff
@@ -191,15 +191,21 @@ export class Popover {
...
+        )[0];
+        if (tooltip) {
+          const mouseLeaveHandler = (() => {
+            const hoverElements = document.querySelectorAll(':hover');
+            const index = [].indexOf.call(hoverElements, tooltip);
+            if (index < 0) {
+              this.hide();
+            }
+          }).bind(this);
+          setTimeout(mouseLeaveHandler, 200);
```
arvindanta, line 207: do we need a settimeout here ?

Vignesh-Manogar-E3433: Yes, we have some props in tooltip/popover that can add distance between trigger and content. So if we don't add the timeout, the hide is getting executed immediately.

arvindanta (approved):

PR #575, opened 2022-04-25
feat(fw-select): added provision to add/update the options dynamically

kishore-kumar-E3682:
## Description:

Fix for the issue raised here - https://community.developers.freshworks.com/t/crayons-unable-to-update-fw-select-options-after-page-load/5555

## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/select/select.tsx
+++ b/packages/crayons-core/src/components/select/select.tsx
@@ -345,6 +345,22 @@ export class Select {
     this.renderInput();
   }
 
+  @Watch('options')
+  onOptionsChange(newValue) {
+    this.dataSource = newValue;
+    const selectedValues = newValue?.filter((option) => option.selected);
+    if (selectedValues.length > 0) {
+      // selected flag is available in the datasource
+      this.selectedOptionsState = selectedValues;
+      this.value = this.multiple
+        ? this.selectedOptionsState[0]?.value
+        : this.selectedOptionsState.map((x) => x.value);
+    } else {
+      this.value = this.multiple ? [] : '';
+      this.selectedOptionsState = [];
+    }
+  }
+
   @Method()
   async getSelectedItem(): Promise<any> {
     return this.fwListOptions.getSelectedOptions();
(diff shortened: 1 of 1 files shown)


packages/crayons-core/src/components/select/select.tsx
```diff
@@ -345,6 +345,22 @@ export class Select {
...
 
+  @Watch('options')
+  onOptionsChange(newValue) {
+    this.dataSource = newValue;
+    const selectedValues = newValue?.filter((option) => option.selected);
+    if (selectedValues.length > 0) {
+      // selected flag is available in the datasource
+      this.selectedOptionsState = selectedValues;
+      this.value = this.multiple
+        ? this.selectedOptionsState[0]?.value
```
Vignesh-Manogar-E3433, line 356: The condition seems to be in reverse.

packages/crayons-core/src/components/select/select.tsx
```diff
@@ -345,6 +345,22 @@ export class Select {
     this.renderInput();
   }
 
+  @Watch('options')
+  onOptionsChange(newValue) {
+    this.dataSource = newValue;
+    const selectedValues = newValue?.filter((option) => option.selected);
```
Vignesh-Manogar-E3433, line 351: This seems to be redundant in componentWillLoad too. Should we separate this out into a private function?

kishore-kumar-E3682: During componentWillLoad the datasource can be either the slot or the option prop, but in case of `@Watch('options')` the datasource is just options and we follow the hierarchy here datasource > value > selectedOptions but at this point setting the value before options wont work because on componentWillLoad this hierarchy is not there.

packages/crayons-core/src/components/select/select.tsx
kishore-kumar-E3682: Thanks for pointing it out!!

Vignesh-Manogar-E3433 (approved):

arvindanta (approved):

PR #604, opened 2022-05-18
fix(data-table): separating isLoading and showShimmer functionality

Vignesh-Manogar-E3433:
In this PR, we are removing the dependency between isLoading and showShimmer props.

Issue:
Table moves to a loading state intermittently. This disabled other actions on the table.

Fix: 
The issue seems to be with setting isLoading on componentWillLoad and removing it only when showShimmer is false. componentWillLoad seems to be called as part of component reattached state too like during HMR. So we removed isLoading from componentWillLoad.

## Checklist:
 
## How Has This Been Tested?
Tested in Chrome. 
Tested locally in CO UI project.

### DIFF
--- a/packages/crayons-core/src/components/data-table/data-table.tsx
+++ b/packages/crayons-core/src/components/data-table/data-table.tsx
@@ -214,7 +214,6 @@ export class DataTable {
    * componentWillLoad lifecycle event
    */
   componentWillLoad() {
-    this.isLoading = true;
     this.columnOrdering(this.columns);
     if (localStorage && this.autoSaveSettings) {
       const tableId = this.el.id ? `-${this.el.id}` : '';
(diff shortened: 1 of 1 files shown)


packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -1606,13 +1604,15 @@ export class DataTable {
...
         shimmerTemplate.push(
           <tr>
-            {[...Array(columnsLength).keys()].map(() => {
-              return (
-                <td>
-                  <fw-skeleton height='12px'></fw-skeleton>
-                </td>
-              );
-            })}
+            {Array(columnsLength)
```
Vignesh-Manogar-E3433, line 1607: Same logic. Just that using 'fill' works better after transpiling.

arvindanta (approved):

PR #611, opened 2022-05-26
feat(fw-timepicker): added placement option for the dropdown

kishore-kumar-E3682:
## Description:
- Exposed caret and optionsPlacement prop to control the position of the dropdown
- Removed the FieldControl as it is used in the fw-select.

## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/timepicker/timepicker.scss
+++ b/packages/crayons-core/src/components/timepicker/timepicker.scss
@@ -1,7 +1,3 @@
-/**
-  @prop --fw-hint-color: Color of the hint text.
-  @prop --fw-warning-color: Color of the warning text.
-  @prop --fw-error-color: Color of the error text.
-*/
-
-@import '../../function-components/field-control.scss';
+:host {
+  display: block;
+}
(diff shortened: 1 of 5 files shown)


packages/crayons-core/src/components/timepicker/timepicker.scss
```diff
@@ -1,7 +1,3 @@
-/**
-  @prop --fw-hint-color: Color of the hint text.
-  @prop --fw-warning-color: Color of the warning text.
-  @prop --fw-error-color: Color of the error text.
-*/
-
-@import '../../function-components/field-control.scss';
+.timepicker {
```
arvindanta, line 1: why do we need a `block` display here ?. It would mean, we cannot have other components inline with this.

kishore-kumar-E3682: Without the `block` element the consuming component can't specify the width or the `margin/padding` for the component.

kishore-kumar-E3682: I can be either a `block` or `inline-block` since all the other components are block elements i have used the same here.

arvindanta (approved):

PR #614, opened 2022-05-30
feat(design tokens): converting design tokens to css variables

kishore-kumar-E3682:
## Description:
- Created new package `crayons-core` to store the system's design-tokens
- The tokens are converter to css variables at build time
- Exposed API's to set the theme by the consuming module at runtime.

## Checklist:


packages/crayons-core/src/index.html
```diff
@@ -12,7 +12,22 @@
     <script nomodule src="/build/crayons.js"></script>
   </head>
   <body>
+    <fw-button expand mode="freshsales" color="primary"
```
arvindanta, line 15: do we need to use `mode` ?.

Kindly comment the components here as this will increase the initial bundle size of the script.

packages/crayons-styles/design-tokens/crayons/base.json
```diff
@@ -0,0 +1,237 @@
...
+        "100": {
+          "value": "#fff",
+          "type": "color",
+          "var": "--fw-color-neutral-100"
+        }
+      }
+    },
+    "font": {
+      "3xl": {
+        "value": " 2.75rem",
```
arvindanta, line 166: I see some space before the value. Is it required ?

packages/crayons-styles/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "homepage": "https://github.com/freshworks/crayons#readme",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/freshworks/crayons.git"
+  },
+  "scripts": {
+    "build:tokens": "node ./scripts/convertTokens.js",
+    "create:types": "npx -p typescript tsc",
```
arvindanta, line 26: can u install typescript as dev dependency here instead of `npx`

packages/crayons-styles/src/index.js
```diff
@@ -0,0 +1,50 @@
+import crayons from '../dist/design-tokens/crayons.json';
```
arvindanta, line 1: let's not point to `dist` folder directly inside `src`.

packages/crayons-styles/src/index.js
```diff
@@ -0,0 +1,50 @@
...
+const base = '$base.';
+
+function getValue(value) {
```
arvindanta, line 4: `getValue` and `convertTokenToString` are getting repeated in 2 files. can u reuse them instead?

packages/crayons-styles/scripts/convertTokens.js
```diff
@@ -0,0 +1,81 @@
+const fs = require('fs');
+const path = require('path');
+
+const tokenOutDir = './dist/design-tokens/';
```
arvindanta, line 4: does this script after creating dist folder by rollup ?.

arvindanta (changes requested):

packages/crayons-styles/scripts/convertTokens.js
kishore-kumar-E3682: No, The scripts runs before the rollup. All the design token sets are combined into a single based on the deisgn system and resulting file is created in dist folder. Which will be picked up by rollup and converted into js files for dynamic import.

packages/crayons-styles/src/index.js
kishore-kumar-E3682: The JSON file is created during build. Yes we can move it to a folder outside dist and add the folder to .gitignore. But this looks simpler.

packages/crayons-styles/design-tokens/crayons/base.json
kishore-kumar-E3682: Nope, it's a typo. Fixed it!!

packages/crayons-core/src/index.html
kishore-kumar-E3682: Reverted the changes to index.html, the changes was there to showcase the demo.

packages/crayons-styles/src/index.js
arvindanta: please move it to a folder outside dist.

packages/crayons-styles/src/index.js
```diff
@@ -0,0 +1,25 @@
...
+  const sheet = new CSSStyleSheet();
+  sheet.replaceSync(`:root {${cssString}}`);
+  document.adoptedStyleSheets = [sheet];
+}
+
+export function setTheme(theme) {
+  if (theme === 'crayons') {
+    setStyle(crayons);
+  }
+  import(`../dist/design-tokens/${theme}.json`)
```
arvindanta, line 15: point to a different folder other than dist here.

kishore-kumar-E3682: Renamed the output folder of the design tokens

arvindanta (approved):

PR #619, opened 2022-06-03
feat(fw-datepicker): added time to single date picker

kishore-kumar-E3682:
BREAKING CHANGE: changed the default value of format in timepicker from `hh:mm a` to locale based format.
`min` and `max` time will be based on locale time format

## Description:
Added provision to choose time in the single date picker.
Screenshot 2022-06-03 at 2 41 29 PM

## Checklist:


packages/crayons-core/src/components/datepicker/datepicker.tsx
```diff
@@ -204,6 +206,18 @@ export class Datepicker {
    * Label displayed on the interface, for the component.
    */
   @Prop() label = '';
+  /**
+   * Whether the time-picker should be shown in the data-picker.
```
kulkarni-kashyap, line 210:
```suggestion
   * Whether the time-picker should be shown in the date-picker.
```

packages/crayons-core/src/components/datepicker/datepicker.tsx
```diff
@@ -204,6 +206,18 @@ export class Datepicker {
...
+   */
+  @Prop() time = false;
+  /**
+   * The props for the time picker.
+   */
+  @Prop() timeProps = {};
+  /**
+   * The format of time picker .
+   */
+  @Prop() timeFormat: 'HH:mm' | 'hh:mm a' = 'hh:mm a';
```
kulkarni-kashyap, line 220: Now, with this attribute being introduced, can we consider renaming existing `displayFormat` to something like `dateDisplayFormat` to reflect that it points to date alone ?

packages/crayons-core/src/components.d.ts
```diff
@@ -331,6 +331,18 @@ export namespace Components {
           * Theme based on which the input of the datepicker is styled.
          */
         "state": 'normal' | 'warning' | 'error';
+        /**
+          * Whether the time-picker should be shown in the date-picker.
+         */
+        "time": boolean;
```
arvindanta, line 337: can you rename this to showTimePicker to ensure it is clear and explicit.

packages/crayons-core/src/components/datepicker/datepicker.tsx
arvindanta: Can we just use the `displayFormat` and allow the user to pass the format with time or just date format alone - displayFormat for timepicker - `yyyy/MM/dd HH:mm`
displayFormat for datepicker -  `yyyy/MM/dd`

packages/crayons-core/src/components/datepicker/readme.md
```diff
@@ -87,6 +87,38 @@ function App() {
 </code-block>
 </code-group>
 
+## Date time picker
+
+Date picker is currently supported only on single date picker.
```
arvindanta, line 92: Change it to `Date time picker`

arvindanta (changes requested):

packages/crayons-core/src/components/datepicker/datepicker.tsx
kishore-kumar-E3682: If we use a single prop for both time and date format the used can use any pattern to enter the date time format and the format need to be split to get the time format and passed to `fw-timepicker`. The drawback with that approach is the end user can use any pattern to set the format like `d-MM-yy HH:mm` or  `d-MM-yy - HH:mm a` or even the worst case scenario `HH:mm - yyyy/MM/dd` in cased like these it difficult to get the time format from the display format. Hence i decided to for different props for date and time format as it easy to validate the too.

arvindanta: Can you please check if `date-fns` supports parsing date-time formats . If that's case, using just a single format will help. `date-fns` util can return the appropriate date and time format.

arvindanta: Also pls check if i18n support works with Timepicker as well. For datepicker the format changes based on the lang. Similarly I would advise we change the Timepicker's format as appropriate.

arvindanta (changes requested):

packages/crayons-core/src/components/datepicker/datepicker.tsx
kishore-kumar-E3682: Added Locale support to Time picker. Now the timepickers format will be based on the locale if the format/timeFormat
 is not provide

packages/crayons-core/src/components.d.ts
```diff
@@ -331,10 +331,22 @@ export namespace Components {
...
         /**
           * Theme based on which the input of the datepicker is styled.
          */
         "state": 'normal' | 'warning' | 'error';
+        /**
+          * The format of time picker .
+         */
+        "timeFormat": string;
+        /**
+          * The props for the time picker. Refer the fw-timepicker for valid format.
```
kulkarni-kashyap, line 347:
```suggestion
          * The props for the time picker. Refer the fw-timepicker for valid props.
```

kulkarni-kashyap (approved):

arvindanta (approved):

PR #627, opened 2022-06-17
fix(rtl): fixing popovers placement when direction is rtl

Vignesh-Manogar-E3433:
Position of the popover modified with popper js custom modifier.

Before Fix:
Screenshot 2022-06-17 at 11 14 34 AM
Screenshot 2022-06-17 at 11 14 50 AM
Screenshot 2022-06-17 at 11 14 22 AM

After Fix:
Screenshot 2022-06-17 at 11 08 10 AM
Screenshot 2022-06-17 at 11 08 59 AM
Screenshot 2022-06-17 at 11 10 42 AM

## Checklist:
 
## How Has This Been Tested?
Tested in Chrome, Firefox and Safari browsers.

### DIFF
--- a/packages/crayons-core/src/components/datepicker/datepicker.tsx
+++ b/packages/crayons-core/src/components/datepicker/datepicker.tsx
@@ -98,6 +98,7 @@ export class Datepicker {
   @State() toMonth: number;
   @State() firstFocusElement: HTMLElement = null;
   @State() lastFocusElement: HTMLElement = null;
+  @State() popoverContentElement: HTMLElement;
   @State() langModule: any;
   @State() shortMonthNames: any;
   @State() longMonthNames: any;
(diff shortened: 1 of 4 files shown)


packages/crayons-core/src/components/datepicker/datepicker.tsx
```diff
@@ -98,6 +98,7 @@ export class Datepicker {
   @State() toMonth: number;
   @State() firstFocusElement: HTMLElement = null;
   @State() lastFocusElement: HTMLElement = null;
+  @State() popoverContentElement: HTMLElement;
```
arvindanta, line 101: should this be a state or can we just have it as a private variable ?

Vignesh-Manogar-E3433: This has to be a state, else boundary is null when select is first rendered.

This is the same as the case in the thread here:
https://stackoverflow.com/questions/59729258/stenciljs-jsx-use-ref-in-as-a-prop

kishore-kumar-E3682 (approved):

PR #638, opened 2022-07-13
feat(fw-timepicker): added locale support

kishore-kumar-E3682:
BREAKING CHANGE: changed the default value of format from hh:mm a to locale based format.

## Checklist:


packages/crayons-core/src/components/timepicker/timepicker.tsx
```diff
@@ -26,22 +34,19 @@ export class Timepicker {
...
    */
   @Prop() disabled = false;
 
-  /**
-   * Represent the intervals and can be a number or array of numbers representing the minute values in an hour
-   */
-  @State() isMeridianFormat?: boolean = this.format === 'hh:mm a';
-
   /**
    * Time output value
```
kulkarni-kashyap, line 51: Are we expecting that value should always be of `HH:mm` format only ?

packages/crayons-core/src/components/timepicker/timepicker.tsx
```diff
@@ -209,6 +206,19 @@ export class Timepicker {
...
+  async handleLocaleChange(newLocale) {
+    this.langModule = await TranslationController.getDateLangModule(newLocale);
+    this.format =
+      this.format || this.langModule?.formatLong?.time({ width: 'short' });
+    this.minTime = isMatch(this.minTime, this.format)
+      ? this.minTime
+      : format(startOfDay(new Date()), this.format);
+    this.maxTime = isMatch(this.maxTime, this.format)
+      ? this.maxTime
+      : format(endOfDay(new Date()), this.format);
```
kulkarni-kashyap, line 219: I think, for format, it expects the lang module should also be passed as argument

kishore-kumar-E3682: The max and min time are used internally, both these values are converted into milliseconds for setting the upper and lower limit. Also the third is an argument to `format` function is optional.

packages/crayons-core/src/components/timepicker/timepicker.tsx
kishore-kumar-E3682: No the value is expected to of same format as that of the `format` or `locale` attributes. If the `value` attribute and one of the option's value in the select matches it will select that value otherwise it wont.

kulkarni-kashyap (approved):

arvindanta (approved):

PR #677, opened 2022-09-07
refactor: add watcher for prop update, add creatable props

rihansiddhi:
## Checklist:


packages/crayons-core/src/components/select/select.tsx
```diff
@@ -231,6 +231,17 @@ export class Select {
    * Describes the select's boundary HTMLElement
    */
   @Prop() boundary: HTMLElement;
+  /**
+   * Props to be passed for creatable select
+   * isCreatable: boolean - If true, select accepts user input as selected option
+   * validateNewOption: (value) => boolean - If passed, this function will determine the error state for every new option entered
+   * formaatCreateLabel: (label) => string - Gets the label for the "create new ..." option in the menu. Is given the current input value.
```
arvindanta, line 238: `is Given the current input value` . I guess there is a typo here.

packages/crayons-core/src/components/select/select.tsx
```diff
@@ -231,6 +231,17 @@ export class Select {
    * Describes the select's boundary HTMLElement
    */
   @Prop() boundary: HTMLElement;
+  /**
+   * Props to be passed for creatable select
+   * isCreatable: boolean - If true, select accepts user input as selected option
```
arvindanta, line 236: can you be more explicit saying , If true, select accepts user input that are not present as options and add them as options ?

arvindanta (changes requested):

packages/crayons-core/src/components/select/select.tsx
```diff
@@ -646,6 +663,45 @@ export class Select {
     else if (this.state === 'warning') return `warning-${this.name}`;
     return null;
   }
+
+  onPaste(e) {
+    // Allow paste only if isCreatable is true
```
arvindanta, line 668: why are we allowing paste only if `isCreatable` is true. Should we enable this `paste` functionality for all the usecases ?

arvindanta (changes requested):

packages/crayons-core/src/components/select/select.tsx
rihansiddhi: @arvindanta paste functionality can contain values which are not part of the predefined options. With search, we will not be able to know the options available to the user. By allowing this only when isCreatable is true, we would allow this functionality to add values which may or may not exist in the options as user might not need this functionality by default and only when user wants to input custom values.

arvindanta (approved):

PR #712, opened 2022-10-31
style: use border variable for select when disabled

rihansiddhi:
## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/select/select.scss
+++ b/packages/crayons-core/src/components/select/select.scss
@@ -162,7 +162,7 @@ $disabled-color: $input-disabled-color;
 
   &.select-disabled {
     color: $disabled-color;
-    border: 1px solid $color-smoke-50;
+    border: var(--fw-select-border, 1px solid $color-smoke-50);
     background-color: $input-disabled-bg !important;
     border-style: solid;
     cursor: not-allowed;


packages/crayons-core/src/components/select/select.scss
```diff
@@ -162,7 +162,7 @@ $disabled-color: $input-disabled-color;
 
   &.select-disabled {
     color: $disabled-color;
-    border: 1px solid $color-smoke-50;
+    border: var(--fw-select-border, 1px solid $color-smoke-50);
```
arvindanta, line 165: input-container also uses this css variable. Kindly check if reusing the same variable will conflict with the other one. https://github.com/freshworks/crayons/pull/712/files#diff-8cf5161c0a39292c9841c92c6768ae90f6c6e710ecd87ee29855dbf62a3b7d8dR74

rihansiddhi: @arvindanta select-disabled will override the border style applied in input-container when disabled. Whatever has been assigned to --fw-select-border variable will be the border style even if select is disabled or not. Before this, when disabled we had been showing the border as "1px solid $color-smoke-50;" regardless of the css variable being passed.

arvindanta (approved):

PR #721, opened 2022-11-16
feat: add option label path and option value path for select and list…

rihansiddhi:
… options

## Checklist:


packages/crayons-core/src/components/options-list/list-options.tsx
```diff
@@ -40,7 +40,9 @@ export class ListOptions {
       const filteredValue =
         value !== ''
           ? dataSource.filter((option) =>
-              option.text.toLowerCase().includes(value.toLowerCase())
+              option?.[this.optionLabelPath]
```
arvindanta, line 43: I don't think you need optional chaining for option. u can change it to 
`option[this.optionLabelPath]?.toLowerCase()

arvindanta: Please make this changes in other places. option should be a valid object since its passed from `filter` and `map`

arvindanta (changes requested):

arvindanta (approved):

Issue #666, opened 2022-08-24
[Enhancement] - fw-popover provision to delay opening the popup


santhoshbalusamy:
**Description of the enhancement:**
need a provision to delay opening the popup

**Describe the solution you'd like**
Similar to the element UI popup, 
https://element.eleme.io/#/en-US/component/popover
there is an option to specify open-delay,

We need a similar kind of experience with crayons.

PR #725, opened 2022-12-01
enhancement(popover): introduced new props to facilitate delay in open/close transition in popover component

mariappan-subramanian:
There was a need to control the popover open/close transition for Marketplace apps website. Since there was no timing control over these transition it created kind of glitch in the interaction & faced usability issues. Thus introduced additional props to control these transition as in element-plus popover component (https://element.eleme.io/#/en-US/component/popover) which we used earlier in the project.

fix: #666

## Checklist:

## How Has This Been Tested?
Local

### DIFF
--- a/packages/crayons-core/src/components.d.ts
+++ b/packages/crayons-core/src/components.d.ts
@@ -988,6 +988,14 @@ export namespace Components {
          */
         "notFoundText": string;
         /**
+          * Key for determining the label for a given option
+         */
+        "optionLabelPath": string;
+        /**
+          * Key for determining the value for a given option
+         */
+        "optionValuePath": string;
+        /**
           * Value corresponding to the option, that is saved  when the form data is saved.
          */
         "options": any[];
@@ -1445,6 +1453,14 @@ export namespace Components {
           * Default option to be shown if the option doesn't match the filterText.
          */
         "notFoundText": string;
+        /**
+          * Key for determining the label for a given option
+         */
+        "optionLabelPath": string;
+        /**
+          * Key for determining the value for a given option
+         */
+        "optionValuePath": string;
         /**
           * The data for the select component, the options will be of type array of fw-select-options.
          */
@@ -3585,6 +3601,14 @@ declare namespace LocalJSX {
          */
         "onFwLoading"?: (event: FwListOptionsCustomEvent<any>) => void;
         /**
+          * Key for determining the label for a given option
+         */
+        "optionLabelPath"?: string;
+        /**
+          * Key for determining the value for a given option
+         */
+        "optionValuePath"?: string;
+        /**
           * Value corresponding to the option, that is saved  when the form data is saved.
          */
         "options"?: any[];
@@ -4050,6 +4074,14 @@ declare namespace LocalJSX {
           * Triggered when the list box comes into focus.
          */
         "onFwFocus"?: (event: FwSelectCustomEvent<any>) => void;
+        /**
+          * Key for determining the label for a given option
+         */
+        "optionLabelPath"?: string;
+        /**
+          * Key for determining the value for a given option
+         */
+        "optionValuePath"?: string;
         /**
           * The data for the select component, the options will be of type array of fw-select-options.
          */
--- a/packages/crayons-core/src/components/options-list/list-options.e2e.ts
+++ b/packages/crayons-core/src/components/options-list/list-options.e2e.ts
@@ -1,5 +1,39 @@
 import { newE2EPage } from '@stencil/core/testing';
 describe('fw-list-options', () => {
+  const props = {
+    options: [
+      {
+        name: 'Angela Smith',
+        subText: 'angela.smith@gmail.com',
+        email: 'angela.smith@gmail.com',
+        graphicsProps: {
+          image:
+            'https://images.unsplash.com/photo-1529778873920-4da4926a72c2?ixlib=rb-1.2.1&auto=format&fit=crop&w=300&q=80',
+        },
+      },
+      {
+        name: 'Freshdesk support from India and Berlin',
+        subText: 'support.india@freshdesk.com',
+        email: 'support.india@freshdesk.com',
+        graphicsProps: {
+          image:
+            'https://images.unsplash.com/photo-1529778873920-4da4926a72c2?ixlib=rb-1.2.1&auto=format&fit=crop&w=300&q=80',
+        },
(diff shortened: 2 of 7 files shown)


packages/crayons-core/src/components/popover/popover.tsx
```diff
@@ -142,9 +154,17 @@ export class Popover {
       this.fwShow.emit();
     }
   }
+  @Method()
+  async show() {
+    clearTimeout(this.timerId);
+    // timeout would be 0 until props is supplied explicit
+    this.timerId = setTimeout(() => {
+      this.triggerShowPopOver();
```
kulkarni-kashyap, line 162: Calling `triggerShowPopOver` with `0`  seconds inside `setTimeout` is not guaranteed to make function call immediately. Can we add a check if `showAfter` is not present and make a function call instead of invoking via setTimeout

packages/crayons-core/src/components/popover/popover.tsx
```diff
@@ -170,6 +190,13 @@ export class Popover {
       this.fwHide.emit();
     }
   }
+  @Method()
+  async hide() {
+    clearTimeout(this.timerId);
```
kulkarni-kashyap, line 195: Can we also consider clearing timer inside disconnectedCallback as well ?

kulkarni-kashyap (changes requested):

packages/crayons-core/src/components/popover/popover.tsx
mariappan-subramanian: Done

packages/crayons-core/src/components/popover/popover.tsx
mariappan-subramanian: Done

packages/crayons-core/src/components/popover/popover.tsx
```diff
@@ -108,7 +120,12 @@ export class Popover {
   }
 
   @Method()
-  async show() {
+  async delayPopOverTrigger(fn, delay) {
```
kulkarni-kashyap, line 123: Can you please elaborate the reason why `delayPopOverTrigger` is exposed as a public method ? , we usually decorate a function with method decorator if it needs to be exposed as a public API.

I think, we can just add one more method in addition to existing `show` method , with name like `showWithDelay` that can accept delay as a param and that can internally call `show` within setTimeout. Similarly for `hide`.

kulkarni-kashyap (changes requested):

packages/crayons-core/src/components/popover/popover.tsx
mariappan-subramanian: Agree, I was not aware of the public API exposing part - Corrected it. 

Regarding the show/hide delay logic, didn't expose any public methods and tried to handle trigger with a handler method to avoid duplication of main logic.

Correct me if this as well can be improved.

kulkarni-kashyap: Thanks for the implementing the changes!

packages/crayons-core/src/components/popover/popover.tsx
```diff
@@ -239,6 +248,32 @@ export class Popover {
...
+  handlePopOverHide() {
+    clearTimeout(this.timerId);
+    if (this.hideAfter > 0) {
+      this.hideWithDelay();
+    } else {
+      this.hide();
+    }
+  }
+
+  async showWithDelay() {
```
kulkarni-kashyap, line 269: Can you please explain why this and the below methods are marked as async ?
Marking something as async will create a promise for us, which is not required for us in this scenario.

kulkarni-kashyap (changes requested):

packages/crayons-core/src/components/popover/popover.tsx
```diff
@@ -183,16 +192,16 @@ export class Popover {
     if (this.trigger === 'click') {
       this.triggerRef.addEventListener(this.trigger, () => {
         if (this.isOpen) {
-          this.hide();
+          this.handlePopOverHide();
```
kulkarni-kashyap, line 195: Can we keep the calls to show and hide here as it is and handle the required changes in `show` and `hide` methods ?

kulkarni-kashyap (changes requested):

packages/crayons-core/src/components/popover/popover.tsx
mariappan-subramanian: This async is not needed anymore, removed it. Got an error insisting to add this earlier while playing around, now looks fine without it. Thanks for pointing out.

packages/crayons-core/src/components/popover/popover.tsx
mariappan-subramanian: Makes sense, corrected it. 
But, not able to handle it in a recursive way as I don't want to introduce an additional flag to use in the breaking condition. Instead added a private method aligning with above suggestion at the same time avoiding duplication of logic.

packages/crayons-core/src/components/popover/popover.tsx
```diff
@@ -86,6 +87,14 @@ export class Popover {
    * Indicates whether popover contents should be hidden on pressing Tab.
    */
   @Prop() hideOnTab = true;
+  /**
+   * Indicates the delay to fire the show popover trigger event
```
kulkarni-kashyap, line 91:
```suggestion
   * Indicates the delay after which popover will be shown.
```

packages/crayons-core/src/components/popover/popover.tsx
```diff
@@ -86,6 +87,14 @@ export class Popover {
...
+   */
+  @Prop() showAfter = 0;
+  /**
+   * Indicates the delay to fire the hide popover trigger event
```
kulkarni-kashyap, line 95:
```suggestion
   * Indicates the delay after which popover will be hidden.
```

kulkarni-kashyap:
@mariappan-subramanian Since there no new dependencies added for the changes required, can we please remove package-lock.json, since the changes here denotes updating some of existing dependencies.

kulkarni-kashyap (changes requested):

packages/crayons-core/src/components/popover/popover.tsx
kulkarni-kashyap: Thanks for accommodating changes!

PR #779, opened 2023-01-19
feat: add kebab menu component, modify row actions in datatable

rihansiddhi:
## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/data-table/data-table.tsx
+++ b/packages/crayons-core/src/components/data-table/data-table.tsx
@@ -84,6 +84,23 @@ export class DataTable {
   @Prop({ mutable: false }) rowActions: DataTableAction[] = [];
 
   /**
+   * To show row actions as a kebab menu
+   */
+  @Prop({ mutable: false }) showRowActionsAsMenu = false;
+
+  /**
+   * Header label for row actions row
+   */
+  @Prop({ mutable: false }) rowActionsHeaderLabel =
+    TranslationController.t('datatable.actions');
+
+  /**
+   * Standard is the default option without any graphics other option is icon which places the icon at the beginning of the row.
+   * The props for the icon are passed as iconName and iconLibrary via the rowActions prop.
+   */
+  @Prop() rowActionsMenuVariant: 'standard' | 'icon' = 'standard';
+
+  /**
    * Rows Array of objects to be displayed in the table.
    */
   @Prop({ mutable: true, reflect: false }) rows: DataTableRow[] = [];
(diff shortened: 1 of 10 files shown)


packages/crayons-core/src/components/kebab-menu/readme.md
```diff
@@ -0,0 +1,174 @@
...
+  iconVariant.addEventListener('fwSelect', (e) => {
+    console.log('icon variant fwSelect event', e.detail);
+  })
+</script>
+```
+
+## Usage
+
+<code-group>
+  <code-block title="HTML">
```
arvindanta, line 65: Please add samples for React

packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -961,6 +977,75 @@ export class DataTable {
     }
   }
 
+  /**
+   * getRowActionMenuProps
+   * @param rowData rowData - complete data of the current row
+   * @returns props for kebab menu
+   */
+  getRowActionMenuProps(rowData: DataTableRow) {
```
arvindanta, line 985: use `graphicProps` to ensure it is extensible.

arvindanta (changes requested):

packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -961,6 +983,70 @@ export class DataTable {
...
+    if (this.showRowActionsAsMenu) {
```
arvindanta, line 992: you can remove the check for `showRowActionsAsMenu` here. Since this is already handled when invoking `renderKebabMenu`

packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -1311,49 +1399,54 @@ export class DataTable {
...
+                                      event.code === 'Enter') &&
+                                    this.performRowAction(action, row)
+                                  }
+                                  onClick={() =>
+                                    this.performRowAction(action, row)
+                                  }
+                                  aria-label={action.name}
+                                >
+                                  {action.iconName ? (
+                                    <fw-icon
```
arvindanta, line 1431: can we implement the usage of graphicProps here as well ?. Some of the row actions buttons might also have a usecase to have icon with color.

arvindanta (changes requested):

packages/crayons-core/src/components/data-table/data-table.tsx
rihansiddhi: Yes we can. So we need to support graphicsProps as well as iconName and iconLibrary which were previously used right?

PR #810, opened 2023-03-01
feat(file-uploader-2): file uploader as part of form

Vignesh-Manogar-E3433:
Adding ability to use fw-file-uploader as part of fw-form.

Checked for:
Rendering in FwForm through schema JSON
Required field
setFocus on File uploader
Validation in form
Initial value setting
Reset form
setField value
form submit serialized value check
Static form

## Checklist:


packages/crayons-core/src/components/file-uploader-2/file-uploader-2.tsx
```diff
@@ -158,8 +158,9 @@ export class FileUploader {
 
   /**
    * errors - errors collection.
+   * Mutable as this can be set from form control too based on form validations.
    */
-  @State() errors: any = [];
+  @Prop({ mutable: true }) errors: any = [];
```
arvindanta, line 163: Can you introduce props like State=‘error’|’warning’|’hint’ 
`Errortext, hinttext, warningtext` props similar to what we have in `input` , `select` `datepicker` etc.?

packages/crayons-core/src/components/form-control/form-control.tsx
```diff
@@ -418,6 +419,41 @@ export class FormControl {
           );
         }
         break;
+      case 'FILES':
+        {
+          const multiple = this.fieldProps?.multiple ? true : false;
+          const errors =
+            this.touched && this.error
```
arvindanta, line 426: In the render function below, we are already showing the error message like `field is required`. 
Do we need to pass the errors and show it inside the file uploader component as well?. 

if we add a prop like `state` and set it to error, we can color the border to `red` .

arvindanta (changes requested):

packages/crayons-core/src/components/file-uploader-2/file-uploader-2.tsx
Vignesh-Manogar-E3433: Changed to `errorText` and `hintText` as suggested. As discussed over call, did not use state prop as it was not valid for file uploader.

packages/crayons-core/src/components/form-control/form-control.tsx
Vignesh-Manogar-E3433: The error messages in render was only for slots. Else, it is usually sent as a prop to the form element, which we have handled in file uploader.

packages/crayons-core/src/components/file-uploader-2/file-uploader-2.tsx
```diff
@@ -379,7 +414,7 @@ export class FileUploader {
         errors.push(this.maxFileSizeError);
       }
     }
-    this.errors = [...this.errors, ...errors];
+    this.errorText = errors.length ? errors[0] : '';
```
arvindanta, line 417: Instead of just showing the first error, can we join the error message together with `,` separated and show it ?

packages/crayons-core/src/components/form/form.tsx
```diff
@@ -451,6 +459,18 @@ export class Form {
...
+    const fileProps = (field: string, multiple: boolean) => {
+      let value: any = [];
+      if (this.values[field]) {
+        if (multiple) {
+          value = this.values[field];
+        } else {
+          value = this.values[field][0] ? [this.values[field][0]] : [];
+        }
+      }
+      return { value };
```
arvindanta, line 472: We are returning `[]` even if the type is not multiple and if the `this.values[field]` is not truthy. Is it expected?

arvindanta (changes requested):

packages/crayons-core/src/components/file-uploader-2/file-uploader-2.tsx
Vignesh-Manogar-E3433: As per design, only one error message is shown here. Showing multiple errors in the same line, leads to scenarios we have not though of before. I think the current implmentation keeps it simple. Kindly reconsider.

packages/crayons-core/src/components/form/form.tsx
Vignesh-Manogar-E3433: Yes, this is expected. To keep the type of returning data same. This way anybody implementing it does not have to worry about what the value of multiple prop is.

arvindanta (approved):

PR #830, opened 2023-03-29
feat(fw-textarea): added maxRows to fw-textarea for dynamic Resizing

kunal-singhh:
## Checklist:
feat(fw-textarea): added maxRows to fw-textarea for dynamic Resizing.

fw-textarea-maxRows

## How Has This Been Tested?

- Tested on local host with both variations with `rows` and `max-rows` and with just `max-rows` prop
  - Eg: When rows=4 and max-rows=10, the default row size will be 4 and as the user interacts with textarea the row size will increase/decrease as the per the number of lines in the content.
  - if no rows is set the default row size is considered as `2` which corresponds to the actual row size of the textarea when no rows prop is passed
  - We also have an optional prop called `max-rows-debounce-timer` which can debounce the events callback to improve performance. Its default value is `200ms`

### DIFF
--- a/packages/crayons-core/src/components.d.ts
+++ b/packages/crayons-core/src/components.d.ts
@@ -2019,6 +2019,14 @@ export namespace Components {
          */
         "label": string;
         /**
+          * Max number of rows the textarea can create when user writes content greater than regular rows.
+         */
+        "maxRows"?: number;
+        /**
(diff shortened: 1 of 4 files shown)


arvindanta (changes requested):

arvindanta (approved):

arvindanta:
LGTM

PR #831, opened 2023-03-29
fix(data-table): saved settings reset only on structural change

Vignesh-Manogar-E3433:
When text inside column prop changes, the table settings are reset.

Solution:
Re-apply existing settings if there is no structural change in the new column definition.

## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/data-table/data-table.tsx
+++ b/packages/crayons-core/src/components/data-table/data-table.tsx
@@ -333,8 +333,53 @@ export class DataTable {
    * @param newColumns recent datatable columns value
    */
   @Watch('columns')
-  columnsChangeHandler(newColumns: DataTableColumn[]) {
+  async columnsChangeHandler(newColumns: DataTableColumn[]) {
+    let hasStructuralChanges = false;
+    const existingColumnSettings = await this.getTableSettings();
     this.columnOrdering(newColumns);
+    const currentColumnSettings = await this.getTableSettings();
+
+    // Observe for structural changes
+    for (const currentColumn in currentColumnSettings) {
+      if (
+        Object.prototype.hasOwnProperty.call(
+          currentColumnSettings,
+          currentColumn
+        )
+      ) {
+        const currentColumnSetting = currentColumnSettings[currentColumn];
+        const oldColumnSetting = existingColumnSettings[currentColumn];
+
+        if (currentColumnSetting && oldColumnSetting) {
+          if (oldColumnSetting.lock !== currentColumnSetting.lock) {
+            hasStructuralChanges = true;
+          }
+        } else {
+          hasStructuralChanges = true;
+        }
+      }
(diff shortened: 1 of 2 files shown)


packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -333,8 +333,53 @@ export class DataTable {
...
   @Watch('columns')
-  columnsChangeHandler(newColumns: DataTableColumn[]) {
+  async columnsChangeHandler(newColumns: DataTableColumn[]) {
+    let hasStructuralChanges = false;
+    const existingColumnSettings = await this.getTableSettings();
     this.columnOrdering(newColumns);
+    const currentColumnSettings = await this.getTableSettings();
+
+    // Observe for structural changes
+    for (const currentColumn in currentColumnSettings) {
```
arvindanta, line 343: You can use `Object.keys` instead of using `for in` and `hasOwnProperty` check.

packages/crayons-core/src/components/data-table/data-table.tsx
```diff
@@ -333,8 +333,53 @@ export class DataTable {
...
+            localStorage.setItem(
+              `fw-table${tableId}`,
+              JSON.stringify(currentColumnSettings)
+            );
+          }
+        } catch (error) {
+          console.log(error.message);
+        }
+      }
+      console.warn(
```
arvindanta, line 379: Should you move this `warn` statement inside the `if(tableId)`.? Since we are updating the localStorage only inside this condition ?

packages/crayons-core/src/components/data-table/data-table.tsx
Vignesh-Manogar-E3433: Yes, but I am failing to see the advantage of suggested alternate solution. Even if I use Object.keys, I still have to loop through.

arvindanta: Yes looping is still required but you can avoid the check for `hasOwnProperty`.

packages/crayons-core/src/components/data-table/data-table.tsx
Vignesh-Manogar-E3433: The `if` condition is only to reset the localStorage. When we entered else block @365 itself, column settings has changed, the warn is just to notify user that visually columns would have changed.

packages/crayons-core/src/components/data-table/data-table.tsx
Vignesh-Manogar-E3433: I am getting Object.keys instead of hasOwnProperty check anyways. Just seems like another way to go about this with no realtime noticeable performance gains.

Besides, for..in still feels like the appropriate method to use in this case. Kindly reconsider.

arvindanta (approved):

PR #832, opened 2023-03-30
fix(fw-daterangepicker): add input validations for date range picker

srivalli-sivaramasankaran:
This PR contains fix for date range picker validation scenarios.

https://monumental-hamster-acb120.netlify.app/components/core/datepicker/#demo

## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/datepicker/datepicker.tsx
+++ b/packages/crayons-core/src/components/datepicker/datepicker.tsx
@@ -882,11 +882,27 @@ export class Datepicker {
     this.year = year;
     this.fromDate = fromDate;
     this.toDate = toDate;
-    this.month = getMonth(
+    const chosenFromMonth = getMonth(
       parse(fromDate, this.displayFormat, new Date(), {
         locale: this.langModule,
       })
     );
+    const chosenToMonth = getMonth(
+      parse(toDate, this.displayFormat, new Date(), {
+        locale: this.langModule,
+      })
+    );
+    if (chosenFromMonth === chosenToMonth && this.toMonth === chosenFromMonth) {
+      this.toMonth = chosenFromMonth;
+      this.month = this.toMonth - 1;
+    } else {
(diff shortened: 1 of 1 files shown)


arvindanta:
Queries:

1. Can the same date be selected in both From and To ?
2. Error message is shown as Please Enter a Valid date or Format. This is used for Da
te picker. Are we going to use the same for Date range picker as well?
3. Reduce the debounce timer to 1000ms since the waiting is clearly visible when error has occurred.
4. For the scenario - Max year as '2022' and invalid value explicitly set as '2023-07-30 to 2023-07-31"
If the from date is changed to `2022-07-30`, the error disappears even though the ToDate is `2023-07-31`.

Screenshot 2023-03-30 at 9 13 20 PM

arvindanta (changes requested):

packages/crayons-core/src/components/datepicker/datepicker.tsx
```diff
@@ -875,6 +833,115 @@ export class Datepicker {
...
+        locale: this.langModule,
+      }) ||
+      year < this.minYear ||
+      year > this.maxYear ||
+      toYear < this.minYear ||
+      toYear > this.maxYear ||
+      !this.isDatewithinRange(parsedFromDate.valueOf(), parsedToDate.valueOf())
+    ) {
+      // Invalid date format
+      this.isDateInvalid = !!val && true;
```
arvindanta, line 879: please remove `&& true` . `!!val` is enough to convert it in to a boolean value.

arvindanta (changes requested):

packages/crayons-core/src/components/datepicker/datepicker.tsx
srivalli-sivaramasankaran: fixed

arvindanta (approved):

PR #845, opened 2023-04-20
fix(datepicker): handle user timezone offset

srivalli-sivaramasankaran:
## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/datepicker/datepicker.tsx
+++ b/packages/crayons-core/src/components/datepicker/datepicker.tsx
@@ -341,7 +341,7 @@ export class Datepicker {
             locale: this.langModule,
           })
         )
-      : formatISO(new Date(value));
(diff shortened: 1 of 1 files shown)


packages/crayons-core/src/components/datepicker/datepicker.tsx
```diff
@@ -341,7 +341,7 @@ export class Datepicker {
             locale: this.langModule,
           })
         )
-      : formatISO(new Date(value));
+      : formatISO(new Date(this.handleUserTimeZoneOffset(value)));
```
arvindanta, line 344: should we not handle Timezone when `displayFormat` is set as well in line 339 ?

arvindanta (changes requested):

arvindanta (approved):

PR #879, opened 2023-07-07
Auto complete missed cases added and setFieldsValue and saved second selection issue fix

abisht1991:
## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/options-list/list-options.tsx
+++ b/packages/crayons-core/src/components/options-list/list-options.tsx
@@ -265,8 +265,10 @@ export class ListOptions {
 
   @Method()
   async setSelectedOptions(options: any[]): Promise<any> {
-    this.selectedOptionsState = options;
-    this.setValue(options);
+    if (options) {
+      this.selectedOptionsState = options;
+      this.setValue(options);
+    }
   }
 
   @Method()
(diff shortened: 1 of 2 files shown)


packages/crayons-core/src/components/options-list/list-options.tsx
```diff
@@ -265,8 +265,10 @@ export class ListOptions {
 
   @Method()
   async setSelectedOptions(options: any[]): Promise<any> {
```
arvindanta, line 267: @rihansiddhi do we have any case where someone can set `setSelectedOptions(null)`.
Can you please validate this ?

rihansiddhi: @arvindanta Yes, we are setting selected options as null using setSelectedOptions method for some cases which will be impacted.

CC @abisht1991

abisht1991: @arvindanta @rihansiddhi 
This would do right, kindly confirm will make the changes.

`if (options || options === null) {
      this.selectedOptionsState = options;
      this.setValue(options);
    }`

rihansiddhi: @abisht1991 Will checking `if (options !== undefined)` suffice your usecase?

abisht1991: yes that would also work @rihansiddhi

arvindanta (approved):

PR #888, opened 2023-09-05
feat(fw-toast): Handles prevent rendering the duplicate toaster multiple times with in the timeout period

madhan028:
## Checklist:


packages/crayons-core/src/components/toast/toast-util.ts
```diff
@@ -93,6 +98,25 @@ export function createToastNotification(
   containerElem.appendChild(toastElem);
 }
 
+export function preventDuplicates(
+  toasteContainerCollection: HTMLCollection,
+  newToasterOption: ToastOptions
+): boolean {
+  let hasDuplicates = false;
+  Array.from(toasteContainerCollection).map((existingToast) => {
```
arvindanta: please use `forEach` instead of `map` since you are not returning any value here.

packages/crayons-core/src/components/toast/toast.tsx
```diff
@@ -49,13 +50,21 @@ export class Toast {
...
+  @Prop() shouldPreventDuplicates: boolean;
+
   componentWillLoad(): void {
     this.toastContainer = createToastStack(this);
   }
 
   @Method()
   async trigger(opts: ToastOptions): Promise<void> {
-    createToastNotification(opts, this.toastContainer, this);
+    const hasDuplicates = preventDuplicates(this.toastContainer.children, opts);
```
arvindanta: @madhan028  can u check if opts.shouldPreventDuplicates is true and then call this method . You can do a short circuit here itself.

arvindanta (changes requested):

packages/crayons-core/src/components/toast/toast-util.ts
```diff
@@ -93,6 +98,25 @@ export function createToastNotification(
   containerElem.appendChild(toastElem);
 }
 
+export function preventDuplicates(
+  toasteContainerCollection: HTMLCollection,
+  newToasterOption: ToastOptions
+): boolean {
+  let hasDuplicates = false;
+  Array.from(toasteContainerCollection).forEach((existingToast) => {
+    if (!newToasterOption.shouldPreventDuplicates) return;
```
sandeepbonala-fw: We can skip the forEach altogether if the flag is false. Bring it to the top/wrap the Array.from.forEach in this if statement, meaning only execute the loop if the flag is true

madhan028: yes sandeep, if you see the below file where this function gets invoked we have handled that logic. I think that covers the idea of not executing this function itself when the flag is turned off. wdyt @sandeepbonala-fw?

arvindanta: @madhan028 You can remove this `if` condition here. since you are already checking for this condition. so inside the `forEach` loop, `shouldPreventDuplicates` will always be `true` . You are doing this check in https://github.com/freshworks/crayons/pull/888/files#diff-ae7c2919363874eb1d6cf1f9dbf528ba39ae06d1507cc5c8d5b84dbd61063d0eR64

arvindanta (approved):

arvindanta:
@madhan028 Great work :)

PR #911, opened 2024-03-14
[FC-120761][FE]:adding sla timer fields in fw-field-editor

Elayaraman:
## Checklist:

## How Has This Been Tested?

This change is made to add new toggle in conversation properies (status)
when we passing a feature flag from the FE
so it will be dynamic according to the value we are passing

image
image

### DIFF
--- a/packages/crayons-extended/custom-objects/src/components/form-builder/components/field-editor.tsx
+++ b/packages/crayons-extended/custom-objects/src/components/form-builder/components/field-editor.tsx
@@ -1322,12 +1322,12 @@ export class FieldEditor {
 
   private renderStatusToggle(objFormValue) {
     const strBaseClassName = 'fw-field-editor';
-    const choices = objFormValue.choices;
-
+    const choices = objFormValue?.choices;
     const isSlaEnabled =
-      objFormValue.field_options.slaTimerEnabled &&
-      (Object(choices[1].choice_options).keys.length > 1 ||
-        Object(choices[2].choice_options).keys.length > 1);
+      objFormValue?.field_options?.slaTimerEnabled &&
+      (choices[1]?.choice_options?.pause_resolution_sla_timer !== undefined ||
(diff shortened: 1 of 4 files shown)


prasanna-vijayan (approved):

packages/crayons-extended/custom-objects/src/components/form-builder/components/field-editor.tsx
```diff
@@ -1352,7 +1364,19 @@ export class FieldEditor {
...
                 </div>
               </span>
-              {toggle}
+              {dataItem?.choice_options?.resolution_timer !== undefined &&
+                renderToggle(
+                  dataItem.id,
+                  'resolution_timer',
+                  dataItem.choice_options.resolution_timer
+                )}
+              {dataItem?.choice_options?.pause_resolution_sla_timer !==
```
rihansiddhi: @Elayaraman Shouldn't we be checking if isSlaEnabled is true here?

Elayaraman: Yeah It should be 
I have made the changes

packages/crayons-extended/custom-objects/src/components/form-builder/components/field-editor.tsx
```diff
@@ -1352,7 +1364,21 @@ export class FieldEditor {
                   ></fw-input>
                 </div>
               </span>
-              {toggle}
+              {isSlaEnabled &&
```
rihansiddhi: @Elayaraman Do we need to have isSlaEnabled check here as this is for resolution_timer?

Elayaraman: No need for it here 
mistakenly added

rihansiddhi (approved):

PR #930, opened 2024-08-05
feat: addition of variation in select options which will let users to add information related to author details

harshith-venkatesh-freshworks:
## Checklist:

## How Has This Been Tested?

I have created a new variation called conversation and have added all the details to create the select options and have added the same in the test case and have also tested the changes locally.
Screenshot 2024-08-05 at 5 11 31 PM


packages/crayons-core/src/components/select-option/select-option.tsx
```diff
@@ -214,7 +227,33 @@ export class SelectOption {
...
+      <span class={'description ' + 'icon-margin'}>{this.text}</span>
+    );
+  }
+
   createIcon() {
+    const { imageSrc } = this.graphicsProps;
+
+    if (imageSrc) {
+      return (
+        <img src={imageSrc} class='image-icon-dimension-standard' alt='icon' />
```
rihansiddhi, line 254: @harshith-venkatesh-freshworks Why do we need an imageSrc here? Can you send it as an svg to data-svg so this can be rendered in fw-icon?

packages/crayons-core/src/components.d.ts
```diff
@@ -1967,6 +1967,10 @@ export namespace Components {
           * Whether clicking on option selects it.
          */
         "allowSelect": boolean;
+        /**
+          * Third line text in conversation can be author details etc.
+         */
+        "author": Author;
```
rihansiddhi: @harshith-venkatesh-freshworks Can the prop name be more generic like metaText or tertiaryText as author is very specific and it can be anything in a conversation item. wdyt?

rihansiddhi (changes requested):

packages/crayons-core/src/components/select-option/select-option.tsx
harshith-venkatesh-freshworks: So, we would be depending on the product here. They have mentioned there would be image source for few icons like whatsapp, instagram icons etc., leading to this approach

packages/crayons-core/src/components.d.ts
harshith-venkatesh-freshworks: @rihansiddhi , sure will go ahead and name it as metaText/tertiaryDetails itself

rihansiddhi (approved):

packages/crayons-core/src/components/select/readme.md
```diff
@@ -1973,7 +1973,7 @@ Refer the [css variables](#css-custom-properties) for modifying the appearance o
...
+| `optionsVariant`      | `options-variant`       | Standard is the default option without any graphics other options are icon and avatar which places either the icon or avatar at the beginning of the row. The props for the icon or avatar are passed as an object via the graphicsProps.                                                                                                                                                                                                                                                                                                                                                                           | `"avatar" \| "conversation" \| "icon" \| "standard"`                                                                                                                 | `'standard'`                                                                                                                       |
```
arvindanta, line 2045: can u add an example in the docs for this `converstation` variant. it will be helpful to easily visualise for someone taking a look at this prop and variants.

harshith-venkatesh-freshworks: Thanks for the review Arvindan, I have added the changes
 for the same

PR #936, opened 2024-09-20
Addition of optional metaText details for Feat/select-options

harshith-venkatesh-freshworks:
## Checklist:

## How Has This Been Tested?

There is an introduction of new variant in previous build where we had introduced new set of data field called metaText.
Here, in case of select-options as well, we are adding the support conditionally to ensure the metaText details if present is made available in the fw-select which is used to create dynamically in fw-form.

Screenshot 2024-09-20 at 2 35 35 PM


harshith-venkatesh-freshworks:
Screenshot 2024-09-23 at 10 31 31 AM


Hi @rihansiddhi @arvindanta , Reverting the above changes, I am still observing the same set of errors here.
How should we be proceeding with the issue?

harshith-venkatesh-freshworks:
Screenshot 2024-09-23 at 5 56 30 PM


Addition of icon alignment for conversation variant

packages/crayons-core/src/components/select-option/select-option.tsx
```diff
@@ -205,6 +205,24 @@ export class SelectOption {
   }
 
   createDescription() {
+    if ((this.subText && this.metaText) || this.metaText) {
```
rihansiddhi: @harshith-venkatesh-freshworks Can we try to unify the method createDescription and createConversationDescription? There are some duplicated codes in these methods which can be further optimised. Also I think we won't be needing the variant as conversation if we allow metaText to be accepted without the variant. We can use variant as icon to display the icon as well as metaText. Can you please check them?

rihansiddhi (changes requested):

packages/crayons-core/src/components/select-option/select-option.tsx
harshith-venkatesh-freshworks: @rihansiddhi , good point, Indeed we can remove createConversationDescription here.
However, for the icon.
Are we able to customise the padding for the display.
I have made some changes to conversation variant.

harshith-venkatesh-freshworks: Let me know your view on the same

packages/crayons-core/src/components/select-option/select-option.tsx
```diff
@@ -188,8 +188,8 @@ export class SelectOption {
         return (
           <Fragment>
             {checkbox}
-            {this.createIcon()}
-            {this.createConversationDescription()}
+            {this.createConversationIcon()}
```
rihansiddhi, line 191: @harshith-venkatesh-freshworks Instead of having a different method for conversation icon, can we add the div with the conversation-icon directly here?
```suggestion
          <div class='conversation-icon'>
            {this.createIcon()}
            </div>
```

rihansiddhi (changes requested):

packages/crayons-core/src/components/select-option/select-option.tsx
harshith-venkatesh-freshworks: Sure Rihan, this is good, thanks for the suggestion

packages/crayons-core/src/components/select-option/select-option.tsx
rihansiddhi:
```suggestion
    if (this.metaText) {
```
@harshith-venkatesh-freshworks Can we simplify this as we are already checking subText condition inside so adding a condition outside wouldn't make any difference right?

rihansiddhi (changes requested):

harshith-venkatesh-freshworks:
Screenshot 2024-09-30 at 3 58 31 PM
Screenshot 2024-09-30 at 3 58 05 PM

packages/crayons-core/src/components/select-option/select-option.e2e.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-debugger */
```
rihansiddhi: Can you remove this? @harshith-venkatesh-freshworks

rihansiddhi (approved):
LGTM. Kindly address the tests in a subsequent PR.

PR #949, opened 2024-11-08
fix: Datepicker emitEvent error handling on parsing of icelandic language

harshith-venkatesh-freshworks:
## Checklist:

### DIFF
--- a/packages/crayons-core/src/components/datepicker/datepicker.tsx
+++ b/packages/crayons-core/src/components/datepicker/datepicker.tsx
@@ -1579,17 +1579,34 @@ export class Datepicker {
           this.value?.split(TranslationController.t('datepicker.to')) || [];
         fromDateStr = fromDateStr?.trim();
         toDateStr = toDateStr?.trim();
-        const parsedFromDate = parse(
-          fromDateStr,
-          this.displayFormat,
-          new Date(),
-          {
+        let parsedFromDate;
+        if (
+          this.langModule?.code === 'is' &&
+          this.dateFormat === 'dd MMM yyyy'
+        ) {
+          parsedFromDate = parseIcelandicDate(
+            fromDateStr,
+            this.langModule
+          ).valueOf();
+        } else {
+          parsedFromDate = parse(fromDateStr, this.displayFormat, new Date(), {
             locale: this.langModule,
-          }
-        ).valueOf();
-        const parsedToDate = parse(toDateStr, this.displayFormat, new Date(), {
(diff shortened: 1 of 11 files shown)


packages/crayons-core/src/components/datepicker/datepicker.tsx
```diff
@@ -999,9 +999,41 @@ export class Datepicker {
 
   processValueChange(val, emitChange = false) {
     // show error if not ISO format and not display format
-    const parsedDate = parse(val, this.displayFormat, new Date(), {
-      locale: this.langModule,
-    });
+    let parsedDate;
+    if (this.langModule?.code === 'is' && this.dateFormat === 'dd MMM yyyy') {
```
rihansiddhi: @harshith-venkatesh-freshworks Can you move this to a common method and the constants separately as we are duplicating the code here?

packages/crayons-core/src/components/datepicker/datepicker.tsx
```diff
@@ -81,6 +81,52 @@ const getWeekDays = (lang): any => {
...
+const parse = (value, displayFormat, date, langModule) => {
+  if (!value) return value;
+  if (langModule?.locale?.code === 'is' && displayFormat === 'dd MMM yyyy') {
+    return parseIcelandicDate(value, langModule);
+  }
+  return parseDate(value, displayFormat, date, langModule);
+};
+
+const isMatch = (value, displayFormat, langModule) => {
+  return (
```
arvindanta: if langmodule is `is`, this will return false. is this the expectation?

harshith-venkatesh-freshworks: Yes @arvindanta , as isMatch is not supported completely for icelandic language dates.
Though, we are parsing through mapping, the isMatch method would still throw error for the same

packages/crayons-core/src/components/datepicker/datepicker.tsx
```diff
@@ -918,10 +937,10 @@ export class Datepicker {
     if (
       !isValidFromDate ||
       !isValidToDate ||
-      !isMatch(fromDate, this.displayFormat, {
+      isNoMatch(fromDate, this.displayFormat, {
```
arvindanta: can u retain isMatch here. and modify the function above ?.

arvindanta (approved):

arvindanta (approved):

rihansiddhi (approved):

PR #963, opened 2025-04-01
Fix section related bugs

Bharatram-Santhanakrishnan:
## Checklist:

 ## Change log

1. `supportDependentAndMSDDInSections` - A new property has been introduced to let users decide whether sectional fields should accept dependent and multiselect dropdown fields. This is to make the section more flexible and at the same time reduce the blast radius in case something goes wrong.

This flag only controls whether the fields are allowed to dropped in a section or not. It does not govern the below mentioned logic.
 
2. The logic for computing the maximum number of fields a section can hold has been modified. Earlier, the logic took into account the `error element` as well into calculation thus returning one more than the actual count. This logic held good  earlier since all the fields had a count of 1. Now that dependent fields are counted as 3, we needed to check what the draggedElement was and its count.

## How Has This Been Tested?

Dev testing has been done by checking the max limit breach for different field types. Checked reordering as well to ensure that the element belonging to the same container does not contribute to the count.

### DIFF
--- a/packages/crayons-core/src/components.d.ts
+++ b/packages/crayons-core/src/components.d.ts
@@ -505,6 +505,10 @@ export namespace Components {
           * Whether the list should be sortable.
          */
         "sortable": boolean;
+        /**
+          * flag to support dependentFields & Multi select dropdown within sections
+         */
+        "supportDependentAndMSDDInSections": boolean;
     }
     interface FwDragItem {
         /**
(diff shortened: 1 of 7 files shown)


prasanna-vijayan (approved):
LGTM

rihansiddhi (approved):

prasanna-vijayan (approved):

PR #965, opened 2025-06-12
Add a flag to prevent section fields reset while using setFieldsValue

Bharatram-Santhanakrishnan:
## Checklist:

## Overview

At present, if we set a section field and it's child fields using setFieldsValue, each update is treated as done by a user and handleInput logic is executed. As per the logic present there, if a section field value is changed, all the child fields are reset. This is a problem because it resets all the child fields that should be updated along with the section field.

## Implementation

1. As a fix, I have introduced a flag that can be passed as an argument to the setFieldsValue method. This would prevent the reset logic from being executed for updates done using this method when the argument (isManualUpdate) is passed.

Since the `setFieldsValue` method and the `handleInput` method are linked by a chain of emitted events, the flag passed would be kept active for a second under the assumption (and based on testing) that the events would be processed within this time period.

2. There is also an enhancement done to the reset logic. Currently, the reset does not clear all the levels of a dependent field within a section. With this change, all the levels of a dependent field within a section is reset properly.

## Impact

1. Impacts `setFieldsValue` and `handleInput` method when called with the new argument. By default, the flag's value defaults to false and hence it won't impact any existing implementations.
2. Reset logic now clears all levels of a dependent field.

### DIFF
--- a/packages/crayons-core/src/components.d.ts
+++ b/packages/crayons-core/src/components.d.ts
@@ -858,7 +858,7 @@ export namespace Components {
         /**
           * Method to set values on the form fields.  param: valuesObj - Object with key as form field name and value as the updated value for the field example: `{ first_name: "new name", last_name: "new last name" }` param: shouldValidate - should this form be validated with the updated values. Default to true.
          */
-        "setFieldsValue": (valuesObj: FormValues, shouldValidate?: boolean) => Promise<void>;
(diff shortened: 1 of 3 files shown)


prasanna-vijayan (approved):
LGTM

rihansiddhi (approved):

gopig93 (approved):

PR #1007, opened 2026-08-10
feat(form-builder): host-driven supportedFieldTypes with per-type limits, DATE_TIME field, async lookup search

sampath-E1005112:
## Checklist:

## How Has This Been Tested?

**Summary**
fw-form-builder can now be told which field types a host application supports, and optionally cap how many of each type can be added, entirely via props - no code change needed on the crayons side to roll a new field type out gradually.

supportedFieldTypes prop (replaces the earlier, never-merged show-multi-select-field interim prop): Array<{ type: string; limit?: number }>. Controls both which field types appear in the "add field" palette / can be composed, and (via the optional limit) overrides the built-in per-type maximum for that type.

// Before (interim, never merged)
show-multi-select-field="false"

// After
supportedFieldTypes={fieldOrder.filter(t => t !== 'MULTI_SELECT').map(type => ({ type }))}
DATE_TIME field type added to CUSTOM_OBJECTS (preset, mapper, limits).

targetSelectProps prop (separate concern, added in a follow-up commit): threads through additional props (e.g. search, debounceTimer) to the target-object fw-select inside fw-fb-field-lookup, so a host can opt a lookup field into server-backed/async search instead of the static option list. Defaults to {}, so existing consumers are unaffected.

**Behavior notes / known limitation**
When a host excludes a type via supportedFieldTypes, any existing field of that type already present in a loaded schema is hidden from the left-nav, field editor, and customize-widget modal - but its data is left untouched in formValues.fields and will still be included on save. This is intentional: the prop controls what's visible/creatable in the builder UI, not a data migration tool. Hosts that need to fully retire a type from a schema should filter/migrate the underlying data themselves.

### DIFF
--- a/packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.e2e.ts
+++ b/packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.e2e.ts
@@ -2005,6 +2005,55 @@ describe('fw-form-builder', () => {
     );
   });
 
+  it('MULTI_SELECT field should not be shown in left nav when showMultiSelectField is false', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(
+      `<fw-form-builder show-multi-select-field=false product-name="CUSTOM_OBJECTS"></fw-form-builder>`
+    );
+
+    const leftPanel = await page.find(
+      'fw-form-builder >>> .form-builder-left-panel'
+    );
+    const fieldItems = await leftPanel.findAll(
+      '.form-builder-left-panel-field-types-list > fw-field-type-menu-item'
+    );
+    expect(fieldItems.length).toEqual(
+      formMapper.CUSTOM_OBJECTS.fieldOrder.length - 2
+    );
+
+    for (const fieldItem of fieldItems) {
+      expect(await fieldItem.getProperty('value')).not.toEqual('MULTI_SELECT');
+    }
+  });
+
+  it('MULTI_SELECT field editor should not be shown when showMultiSelectField is false', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(
+      `<fw-form-builder show-multi-select-field=false product-name="CUSTOM_OBJECTS"></fw-form-builder>`
+    );
+    await page.waitForChanges();
+    await page.$eval(
+      'fw-form-builder',
+      (elm: any, { formValues }: any) => {
+        elm.formValues = formValues;
+      },
+      { formValues: formValues.CUSTOM_OBJECTS }
+    );
+    await page.waitForChanges();
+
+    const rightPanel = await page.find(
+      'fw-form-builder >>> .form-builder-right-panel-field-editor-list'
+    );
+    const fieldDragDropItems = await rightPanel.findAll(
+      'fb-field-drag-drop-item'
+    );
+    expect(fieldDragDropItems.length).toEqual(
+      formValues.CUSTOM_OBJECTS.fields.length - 1
+    );
+  });
+
   it('disables left panel and displays message and button when role is trial, clicking on Explore plan button triggers fwExplorePlan event', async () => {
     const page = await newE2EPage();
 
--- a/packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.e2e.ts
+++ b/packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.e2e.ts
@@ -2054,6 +2054,101 @@ describe('fw-form-builder', () => {
     );
   });
 
+  it('DATE_TIME field should not be shown in left nav when showDateTimeField is false', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(
+      `<fw-form-builder show-date-time-field=false product-name="CUSTOM_OBJECTS"></fw-form-builder>`
+    );
+
+    const leftPanel = await page.find(
+      'fw-form-builder >>> .form-builder-left-panel'
+    );
+    const fieldItems = await leftPanel.findAll(
+      '.form-builder-left-panel-field-types-list > fw-field-type-menu-item'
+    );
+    expect(fieldItems.length).toEqual(
+      formMapper.CUSTOM_OBJECTS.fieldOrder.length - 1
+    );
+
+    for (const fieldItem of fieldItems) {
+      expect(await fieldItem.getProperty('value')).not.toEqual('DATE_TIME');
+    }
+  });
+
+  it('DATE_TIME field should be shown in left nav after Date when showDateTimeField is true', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(
+      `<fw-form-builder show-date-time-field=true product-name="CUSTOM_OBJECTS"></fw-form-builder>`
+    );
+
+    const leftPanel = await page.find(
+      'fw-form-builder >>> .form-builder-left-panel'
+    );
+    const fieldItems = await leftPanel.findAll(
+      '.form-builder-left-panel-field-types-list > fw-field-type-menu-item'
+    );
+    expect(fieldItems.length).toEqual(
+      formMapper.CUSTOM_OBJECTS.fieldOrder.length
+    );
+
+    const fieldTypes = await Promise.all(
+      fieldItems.map(async (fieldItem) => fieldItem.getProperty('value'))
+    );
+    const dateIndex = fieldTypes.indexOf('DATE');
+    const dateTimeIndex = fieldTypes.indexOf('DATE_TIME');
+    expect(dateIndex).toBeGreaterThanOrEqual(0);
+    expect(dateTimeIndex).toBe(dateIndex + 1);
+
+    const dateTimeItem = fieldItems[dateTimeIndex];
+    expect(await dateTimeItem.getProperty('iconName')).toEqual('calendar');
+    expect(await dateTimeItem.getProperty('label')).toEqual('fieldTypeDateTime');
+  });
+
+  it('DATE_TIME field editor should not be shown when showDateTimeField is false', async () => {
+    const page = await newE2EPage();
+    const formValuesWithDateTime = {
+      ...formValues.CUSTOM_OBJECTS,
+      fields: [
+        ...formValues.CUSTOM_OBJECTS.fields,
+        {
+          id: 'a87ce74f-07f9-406a-bd18-bef23a5306a1',
+          name: 'datetime',
+          label: 'Date-Time',
+          type: 'DATE_TIME',
+          required: false,
+          field_options: {},
+          filterable: true,
+          searchable: false,
+        },
+      ],
+    };
+
+    await page.setContent(
+      `<fw-form-builder show-date-time-field=false product-name="CUSTOM_OBJECTS"></fw-form-builder>`
+    );
+    await page.waitForChanges();
+    await page.$eval(
+      'fw-form-builder',
+      (elm: any, { formValues }: any) => {
+        elm.formValues = formValues;
+      },
+      { formValues: formValuesWithDateTime }
+    );
+    await page.waitForChanges();
+
+    const rightPanel = await page.find(
+      'fw-form-builder >>> .form-builder-right-panel-field-editor-list'
+    );
+    const fieldDragDropItems = await rightPanel.findAll(
+      'fb-field-drag-drop-item'
+    );
+    expect(fieldDragDropItems.length).toEqual(
+      formValuesWithDateTime.fields.length - 1
+    );
+  });
+
   it('disables left panel and displays message and button when role is trial, clicking on Explore plan button triggers fwExplorePlan event', async () => {
     const page = await newE2EPage();
 
--- a/packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.e2e.ts
+++ b/packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.e2e.ts
@@ -2036,8 +2036,11 @@ describe('fw-form-builder', () => {
     const fieldItems = await leftPanel.findAll(
       '.form-builder-left-panel-field-types-list > fw-field-type-menu-item'
     );
+    // -2, not -1: renderFieldTypeElement always hides PRIMARY from this
+    // "add field" palette (it's implicit, non-addable) on top of the
+    // MULTI_SELECT exclusion from supportedFieldTypes.
     expect(fieldItems.length).toEqual(
-      formMapper.CUSTOM_OBJECTS.fieldOrder.length - 1
+      formMapper.CUSTOM_OBJECTS.fieldOrder.length - 2
     );
 
     for (const fieldItem of fieldItems) {
@@ -2097,8 +2100,10 @@ describe('fw-form-builder', () => {
     const fieldItems = await leftPanel.findAll(
       '.form-builder-left-panel-field-types-list > fw-field-type-menu-item'
     );
+    // -2, not -1: same PRIMARY carve-out as the MULTI_SELECT exclusion test
+    // above, on top of the DATE_TIME exclusion from supportedFieldTypes.
     expect(fieldItems.length).toEqual(
-      formMapper.CUSTOM_OBJECTS.fieldOrder.length - 1
+      formMapper.CUSTOM_OBJECTS.fieldOrder.length - 2
     );
 
     for (const fieldItem of fieldItems) {
@@ -2119,8 +2124,11 @@ describe('fw-form-builder', () => {
     const fieldItems = await leftPanel.findAll(
       '.form-builder-left-panel-field-types-list > fw-field-type-menu-item'
     );
+    // -1: renderFieldTypeElement always hides PRIMARY from this "add
+    // field" palette (it's implicit, non-addable), independent of
+    // supportedFieldTypes - nothing is excluded here otherwise.
     expect(fieldItems.length).toEqual(
-      formMapper.CUSTOM_OBJECTS.fieldOrder.length
+      formMapper.CUSTOM_OBJECTS.fieldOrder.length - 1
     );
 
     const fieldTypes = await Promise.all(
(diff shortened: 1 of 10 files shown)


packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
```diff
@@ -1429,6 +1436,9 @@ export class FormBuilder {
       return null;
     }
     const strFieldType = dataItem.type;
+    if (strFieldType === 'MULTI_SELECT' && !this.showMultiSelectField) {
```
srivalli-sivaramasankaran: **P0-1 — Possible reorder index mismatch when hidden MULTI_SELECT fields exist in loaded schemas**

**Context (from this PR):** This guard removes existing `MULTI_SELECT` fields from the DOM (`return null`) while they remain in `localFormValues.fields`.

**Concern:** Visible `fb-field-drag-drop-item` elements keep original array `index` props, but `droppedIndex` in `fieldTypeDropHandler` / `fwRepositionField.emit` (~line 695) is computed from visible DOM children only (`draggable.ts`). If a hidden `MULTI_SELECT` sits mid-list, reorder may emit a `targetIndex` that maps to the hidden slot.

**Goal-dependent — please confirm intent:**
- **Greenfield only** (no existing multi-select in schema) → likely fine.
- **Legacy schemas** with existing `MULTI_SELECT` + drag-and-drop reorder → likely needs a fix or documented limitation.

srivalli-sivaramasankaran: **P1-1 — Existing MULTI_SELECT fields become invisible but stay in formValues**

**Context (from this PR):** Same guard — hides the editor for existing `MULTI_SELECT` when the flag is `false`.

**Concern:** Field data remains in `localFormValues.fields` and will still be included on save, but users cannot edit, delete, or reorder via the UI. Note: `showLookupField=false` still renders existing relationship editors — this prop is more aggressive.

**Goal-dependent — please confirm intent:**
- **Hide new creation only** → may be acceptable.
- **Fully unsupported product** with legacy schemas → consider read-only/disabled UI (with delete?) or document that hosts must filter on save.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
```diff
@@ -1546,6 +1556,12 @@ export class FormBuilder {
         arrFieldOrder.splice(dependentIndex, 1);
       }
     }
+    if (!this.showMultiSelectField) {
```
srivalli-sivaramasankaran: **P1-2 — Widget customize modal may still expose MULTI_SELECT (unchanged path, triggered by partial hide in this PR)**

**Context (from this PR):** Multi-select is hidden in left nav + field editor, but `renderWidgetElement` (~line 1504) and `openCustomizeWidgetModalHandler` (~line 830) were not updated. `CUSTOM_OBJECTS` has `"customizeWidget": true`.

**Concern:** Users may still see/select `MULTI_SELECT` in the customize-widget modal, and `fwSaveWidgetFields` may include those IDs — inconsistent with the hide goal elsewhere.

**Goal-dependent — please confirm intent:**
- **Hide from field builder only** → may be OK to leave.
- **Fully invisible when flag is false** → guard `renderWidgetElement` + filter in `openCustomizeWidgetModalHandler`.

**Your call:** In scope for this PR, or follow-up?

srivalli-sivaramasankaran: **P1-3 — Search empty-state may break for hidden MULTI_SELECT matches (downstream of render guard in this PR)**

**Context (from this PR):** `searchChangeHandler` (~738) is unchanged and can still match `MULTI_SELECT` fields (e.g. label "People"). But `renderFieldEditorElement` now returns `null` for them, so `fieldElements` can be `[null]`. `boolShowEmptySearchResults` (~line 1610) checks array length, not rendered count → blank panel instead of "no results".

**Goal-dependent — please confirm intent:**
- Schemas **without** multi-select → low priority.
- **Legacy schemas** + search enabled → fix empty-state (`fieldElements.filter(Boolean).length`) and/or exclude hidden types in search.

**Your call:** Known limitation, fix in this PR, or follow-up?

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
```diff
@@ -1429,6 +1465,9 @@ export class FormBuilder {
       return null;
     }
     const strFieldType = dataItem.type;
+    if (!this.isFieldTypeAllowed(strFieldType)) {
```
srivalli-sivaramasankaran: **Follow-up (P0-1) — still open after `supportedFieldTypes` refactor**

The generalized `isFieldTypeAllowed()` guard here still hides disallowed fields from the DOM while they remain in `localFormValues.fields`. Visible `fb-field-drag-drop-item` elements keep original array `index` props, but `fwRepositionField` (~line 724) still emits DOM-based `droppedIndex` without remapping.

**Still applies** when a host excludes a type (e.g. `MULTI_SELECT`) that already exists mid-list in a loaded schema. E2e fixtures place excluded types last, so this gap is not covered by tests.

**Suggested follow-up:** remap indices before emit, or filter+remap the rendered list; add e2e with an excluded type at a non-last index.

srivalli-sivaramasankaran: **Follow-up (P1-1) — still open**

Excluding a type via `supportedFieldTypes` still hides existing field editors (`return null`) while the data remains in `formValues` — no edit/delete/reorder UI. This is now generalized beyond `MULTI_SELECT` (also affects `DATE_TIME` and any future excluded type).

If the intended contract is "hide creation only" for greenfield schemas, please document. If legacy schemas are in scope, consider read-only/disabled rendering (with delete?) or host-side filtering on save.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
```diff
@@ -1492,7 +1532,7 @@ export class FormBuilder {
   }
 
   private renderWidgetElement(dataItem, intIndex) {
```
srivalli-sivaramasankaran, line 1574: **Follow-up (P1-2) — still open**

`renderFieldEditorElement` and left nav now respect `isFieldTypeAllowed()`, but this method was not updated. For `CUSTOM_OBJECTS` (`customizeWidget: true`), excluded types (e.g. `MULTI_SELECT`) can still appear in the customize-widget modal. `openCustomizeWidgetModalHandler` (~line 859) also still collects all field IDs into `arrWidgetFields`.

**Your call:** In scope for this PR, follow-up, or accepted known limitation?

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
```diff
@@ -1533,19 +1573,20 @@ export class FormBuilder {
...
-      const relationshipIndex = arrFieldOrder.indexOf('RELATIONSHIP');
-      if (relationshipIndex > -1) {
-        arrFieldOrder.splice(relationshipIndex, 1);
-      }
-    }
-    if (!this.showDependentField) {
-      const dependentIndex = arrFieldOrder.indexOf('DEPENDENT_FIELD');
-      if (dependentIndex > -1) {
-        arrFieldOrder.splice(dependentIndex, 1);
+    const arrFieldOrder = [...(objProductPreset?.fieldOrder ?? [])].filter(
```
srivalli-sivaramasankaran, line 1623: **Follow-up (P1-3) — still open**

Left nav and field editor now filter via `isFieldTypeAllowed()`, but `searchChangeHandler` (~line 774) and `boolShowEmptySearchResults` (~line 1635) were not updated. Search can still match excluded types; `fieldElements` may be `[null]` while `fieldElements.length > 0`, so the empty-search UI is not shown — users see a blank panel instead of `noSearchItemsFound`.

**Your call:** Known limitation, fix (`fieldElements.filter(Boolean).length`), and/or exclude disallowed types in search — follow-up?

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
```diff
@@ -98,6 +108,13 @@ export class FormBuilder {
    * flag to show dependentField for CONVERSATION_PROPERTIES or not
    */
   @Prop({ mutable: true }) showDependentField = true;
+  /**
+   * Host-driven list of supported field types and optional per-type limits
+   */
+  @Prop({ mutable: true }) supportedFieldTypes: Array<{
```
srivalli-sivaramasankaran, line 129: **Follow-up — API evolved since initial review**

`showMultiSelectField` from the first commit is now replaced by this `supportedFieldTypes` prop. Functionally covers the original goal (exclude `MULTI_SELECT` via host-provided type list) and adds optional per-type `limit`.

**Please update PR title/description** so reviewers/consumers are not looking for `show-multi-select-field`. Migration example:

\`\`\`js
// Before (interim, never merged)
show-multi-select-field="false"

// After
supportedFieldTypes={fieldOrder.filter(t => t !== 'MULTI_SELECT').map(type => ({ type }))}
\`\`\`

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
srivalli-sivaramasankaran, line 1623: **Follow-up — nice fix**

Copying `fieldOrder` and filtering in place (instead of mutating the shared imported `formMapper` object via `splice`) addresses the test-flakiness / multi-instance concern from the earlier review. 👍

packages/crayons-extended/custom-objects/src/components/form-builder/utils/form-builder-utils.ts
```diff
@@ -129,6 +134,54 @@ export function getMaximumLimitsConfig(productName = 'CUSTOM_OBJECTS') {
   return null;
 }
 
+export function mergeHostFieldLimits(
```
srivalli-sivaramasankaran, line 137: **Follow-up — per-type `limit` has unit tests only**

`mergeHostFieldLimits` / `buildAllowedFieldTypesConfig` are covered in `form-builder-utils.spec.ts`, but there is no e2e verifying that a host-provided `limit` (e.g. `{ type: 'TEXT', limit: 5 }`) disables the left-nav item once the count is reached. Consider adding if this is a consumer-facing guarantee.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.e2e.ts
```diff
@@ -2005,6 +2005,184 @@ describe('fw-form-builder', () => {
...
+    await page.waitForChanges();
+
+    const leftPanel = await page.find(
+      'fw-form-builder >>> .form-builder-left-panel'
+    );
+    const fieldItems = await leftPanel.findAll(
+      '.form-builder-left-panel-field-types-list > fw-field-type-menu-item'
+    );
+    expect(fieldItems.length).toEqual(
+      formMapper.CUSTOM_OBJECTS.fieldOrder.length - 1
```
srivalli-sivaramasankaran: **Follow-up — left-nav count assertion may be off by one**

This expects `fieldOrder.length - 1` when only `MULTI_SELECT` is excluded, but `PRIMARY` is still in `supportedFieldTypes` and `renderFieldTypeElement` returns `null` for `PRIMARY` — visible menu items are likely `length - 2` (PRIMARY + MULTI_SELECT). The initial review used `- 2` for the `show-multi-select-field` test.

Please verify this assertion against actual rendered `fw-field-type-menu-item` count (same note applies to the DATE_TIME exclusion test at ~line 2101).

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
```diff
@@ -86,6 +90,12 @@ export class FormBuilder {
    * object to store the lookup target entities
    */
   @Prop({ mutable: true }) lookupTargetObjects = null;
+  /**
+   * Additional props (e.g. search, debounceTimer) passed through to the target-object fw-select inside
+   * fw-fb-field-lookup, for opting in to server-backed/async search on the target picker. Defaults to an
+   * empty object, so existing consumers who don't set this see no behavior change.
+   */
+  @Prop({ mutable: true }) targetSelectProps = {};
```
srivalli-sivaramasankaran, line 106: **Follow-up — new scope in follow-up commit (`930e0600`)**

`targetSelectProps` threads async/API search through to `fw-fb-field-lookup` → `fw-select`. Default `{}` preserves existing behavior; new `fb-field-lookup.e2e.ts` looks reasonable.

Note: this is a separate concern from `supportedFieldTypes` visibility — consider calling out in PR description or splitting for easier review/rollback.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
sampath-E1005112: Fixed. Extracted buildVisibleFieldIndexMap / resolveDroppedIndex into form-builder-utils.ts and wired them into fieldTypeDropHandler: the DOM-ordinal droppedIndex from draggable.ts is now remapped back to the real index in arrFieldElements before it's used for composeNewField/fwRepositionField, so a hidden (excluded-type) field mid-list no longer shifts the target index for anything landing on the main fieldsContainer (section drops are untouched — they never render excluded top-level types). Added unit tests in form-builder-utils.spec.ts covering an excluded type sitting mid-list (not just at the end), since the existing e2e fixtures only ever place it last. Couldn't add a matching e2e (drag-drop) test — this sandbox can't run any Puppeteer e2e test at all (crashes on the first file jest touches, confirmed environment-wide, not specific to this PR) — so the fix is verified via these unit tests + a plain-logic simulation against the real formMapper data instead.

sampath-E1005112: Keeping today's behavior intentionally rather than changing it: when a host excludes a type via supportedFieldTypes, existing fields of that type are fully hidden from the left-nav/editor/customize-widget modal (not rendered read-only), while their data stays untouched in formValues.fields and is still included on save. supportedFieldTypes controls what's visible/creatable in this builder session, not a data-migration tool — a host that needs to fully retire a type is expected to filter/migrate the underlying schema itself. Documented this explicitly in the prop's doc comment so it's not ambiguous going forward.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
sampath-E1005112: Fixed — renderWidgetElement and the candidate-collection loop in openCustomizeWidgetModalHandler now both respect isFieldTypeAllowed() (PRIMARY excepted, same carve-out as the field editor). An excluded type is no longer offered as a new customize-widget candidate and won't render in the modal list. Fields already saved as widget fields via customizeWidgetFields are left untouched, consistent with the P1-1 decision above.

sampath-E1005112: Fixed — boolShowEmptySearchResults now checks the actually-rendered element count (fieldElements.filter(Boolean).length) instead of the raw array length, so a search/filter match on an excluded type no longer leaves a blank panel — it correctly falls through to noSearchItemsFound.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
sampath-E1005112: Agreed it's a separate concern — called it out as its own section in the updated PR description rather than splitting the PR at this point, since it's already in this branch's history. Happy to split it out if you'd still prefer that.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.e2e.ts
sampath-E1005112: Confirmed, and worse than flagged — verified with a plain-logic simulation against the real fieldOrder data rather than eyeballing it. renderFieldTypeElement always hides PRIMARY from this palette regardless of supportedFieldTypes, so the correct count is fieldOrder.length - 2 (PRIMARY + the excluded type), not - 1. Fixed both the MULTI_SELECT and DATE_TIME exclusion assertions. Also found the identical bug in the adjacent baseline test ("DATE_TIME shown after Date"), which asserted fieldOrder.length with no subtraction at all — fixed to - 1.

packages/crayons-extended/custom-objects/src/components/form-builder/utils/form-builder-utils.ts
sampath-E1005112: Added disables the left-nav item once a host-provided per-type limit is reached in form-builder.e2e.ts. Couldn't execute it locally (this sandbox can't run any Puppeteer e2e — see the P0-1 reply above), so please double check it in CI before merging; logic was hand-verified against validateFormValues/renderFieldTypeElement.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
sampath-E1005112: Done

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
srivalli-sivaramasankaran: **Follow-up review — verified resolved ✅**

`buildVisibleFieldIndexMap` / `resolveDroppedIndex` correctly remap DOM ordinals when excluded types sit mid-list. Unit tests cover the P0-1 scenario. Good fix.

srivalli-sivaramasankaran: **Follow-up review — acknowledged ✅**

Intentional contract is clear from PR body and prop JSDoc. Hosts excluding types must filter/migrate underlying schema on save. No further code change needed from our side.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
srivalli-sivaramasankaran: **Follow-up review — verified resolved ✅**

Widget modal now respects the same allowlist as field editor and left nav; `openCustomizeWidgetModalHandler` also filters new candidates. Good fix.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
srivalli-sivaramasankaran: **Follow-up review — verified resolved ✅**

Empty-search state now uses rendered element count. Search source still includes excluded matches, but UX correctly shows "no results." Good fix.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
```diff
@@ -629,7 +680,16 @@ export class FormBuilder {
     this.removeFieldReorderClass();
     const objDetail = event.detail;
     const elFieldType = objDetail.droppedElement;
-    const intDroppedIndex = objDetail.droppedIndex;
+    let intDroppedIndex = objDetail.droppedIndex;
+    // Only the main fields list can have hidden (isFieldTypeAllowed-excluded)
+    // entries interspersed, so only remap drops landing there - a section's
+    // own drag container never renders excluded top-level field types.
+    if (objDetail.dropToId?.includes('fieldsContainer')) {
+      intDroppedIndex = resolveDroppedIndex(
```
srivalli-sivaramasankaran, line 693: **Follow-up review — P0-1 verified resolved ✅**

`buildVisibleFieldIndexMap` / `resolveDroppedIndex` correctly remap DOM ordinals when excluded types sit mid-list. Unit tests cover this scenario.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
```diff
@@ -98,6 +116,20 @@ export class FormBuilder {
...
+   * Host-driven list of supported field types and optional per-type limits.
+   * Governs what's visible/creatable in this builder session only: a type
+   * left out is hidden from the left-nav, field editor, and
+   * customize-widget modal, including any field of that type already
+   * present in a loaded schema. Excluded fields' data is left untouched in
+   * formValues.fields and will still be included on save - this prop is
+   * not a data migration tool, so a host that needs to fully retire a type
+   * must filter/migrate the underlying schema itself.
+   */
+  @Prop({ mutable: true }) supportedFieldTypes: Array<{
```
srivalli-sivaramasankaran, line 129: **Follow-up review — P1-1 acknowledged ✅**

Intentional contract is clear from PR body and prop JSDoc. Hosts excluding types must filter/migrate underlying schema on save.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
```diff
@@ -1492,7 +1572,14 @@ export class FormBuilder {
   }
 
   private renderWidgetElement(dataItem, intIndex) {
```
srivalli-sivaramasankaran, line 1579: **Follow-up review — P1-2 verified resolved ✅**

Widget modal now respects the same allowlist as field editor and left nav; `openCustomizeWidgetModalHandler` also filters new candidates.

packages/crayons-extended/custom-objects/src/components/form-builder/form-builder.tsx
```diff
@@ -1591,9 +1679,23 @@ export class FormBuilder {
...
+    // consumed by resolveDroppedIndex() in fieldTypeDropHandler.
+    const arrRenderedFieldElements = fieldElements
+      ? fieldElements.filter((element) => element != null)
+      : [];
+    this.visibleFieldIndexMap = buildVisibleFieldIndexMap(
+      arrFieldElements,
+      (fieldType) => this.isFieldTypeAllowed(fieldType)
+    );
+
     const boolShowEmptySearchResults =
```
srivalli-sivaramasankaran, line 1701: **Follow-up review — P1-3 verified resolved ✅**

Empty-search state now uses rendered element count. Excluded search matches correctly show "no results."

srivalli-sivaramasankaran (approved):

